PR #5, opened 2020-03-10
doc: add tools for API generation

redeboer:


redeboer:
@spflueger Please check out locally first and `make html` from `doc`, because there seem to be problems generating documentation for [`tensorwaves/expertsystem/amplitude/canonicaldecay.py`](https://github.com/ComPWA/tensorwaves/blob/9dfd154dd16387a41e3e8f1638f718cf9fa8d6cd/tensorwaves/expertsystem/amplitude/canonicaldecay.py).

doc/Makefile
```diff
@@ -17,4 +17,4 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
-	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O) -W
```
redeboer: I think warnings should be treated as errors: if you decide to write a docstring, the docstring syntax should be strictly correct

doc/.gitignore
```diff
@@ -1,2 +1 @@
-*.rst
```
redeboer: API `rst` files are currently generated automatically, so you want to keep them uncommited.

doc/conf.py
```diff
@@ -8,7 +8,7 @@
 
 # Generate indices for the Python modules in the folder above
 subprocess.call(
-    'sphinx-apidoc -f -M -o api ..',
+    'sphinx-apidoc -f -M -o api ../tensorwaves ../tensorwaves/*constraint*',
```
redeboer: Agree, but why the `../tensorwaves/*constraint*`?

doc/.gitignore
spflueger: but thats how you create documentation in sphinx. that is an absolute must that is removed...

spflueger: i can add api here to filter the generated api out

doc/Makefile
spflueger: Thats true, but I'm afraid I will have to do more fixing then

doc/conf.py
spflueger: Because thats actually a third party module that i copied in (instead of using pip) because I'm not sure about the future. Since this module has a lot of doc string problems I do not want any docstrings generated for this and this statement excludes it

doc/.gitignore
redeboer: Yes I would use `api/*.rst` instead. The files in that folder are generated by `sphinx-api`.

spflueger: I used `api/` for now

Issue #20, opened 2020-04-17
Standardise the local development environment


redeboer:
- [ ] Fix current `requirements.txt` (currently misses e.g. `pyyaml`)
- [x] Add conda environment config file
- [x] Split up `requirements-dev.txt`
- [x] Add pre-commit
- [x] Add flake8 configuration (tox.ini)
- [x] Add pylint config

See also:
https://github.com/ComPWA/pycompwa/pull/63
https://github.com/ComPWA/pycompwa/pull/77
https://github.com/ComPWA/pycompwa/pull/83

This is related to https://github.com/ComPWA/tensorwaves/issues/19

redeboer:
@spflueger 
[This](https://github.com/ComPWA/tensorwaves/pull/22/files#diff-b4ef698db8ca845e5845c4618278f29aR4) may turn out to be problematic, could you replace with `tensorwaves>=2` and try

```bash
conda env create --file environment.yml
```

locally?

Issue #19, opened 2020-04-17
Setup CI


spflueger:
The setup is analogous to the one of pycomwa. It includes

- [x] testing the installation (fairly simple with pure python now)
- [x] testing code via unittests
- [x] code cov
- [x] code quality tests
- [x] building the docs -> deploy to readthedoc (?) on master branch builds
- [x] setup deploy to pypi on release tag

**[Estimated effort: 1-3 days]**

*Additional problems:*
- [x] Fix import of `particle_list.xml` (see [this failed job](https://travis-ci.com/github/redeboer/tensorwaves/jobs/321191484#L450))

redeboer:
Codecov still missing in pull requests

PR #22, opened 2020-04-17
ci: standardize local development environment

redeboer:
Closes https://github.com/ComPWA/tensorwaves/issues/20


requirements.txt
```diff
@@ -1,5 +1,5 @@
 numpy
 progress
-python>=3.6
-tensorflow>=2.0
```
spflueger, line 4: we need tf2

tox.ini
```diff
@@ -0,0 +1,26 @@
...
+[tox]
+skipsdist = True
+envlist =
+    docs,
+    pep8,
+
+[testenv:pep8]
+passenv = PYTHONPATH
+deps =
+    -rrequirements-dev.txt
```
spflueger, line 14: shouldn't this be just `requirements.txt` and `tests/requirements.txt`

redeboer: Hm.. What is needed here is `flake8` and its plugins. They are defined in `requirements-dev.txt`, which also points to the docs and tests requirements because it is supposed to be the complete requirements to do development on tensorwaves.

redeboer:
> Can you also please remove the AmpliTF submodule, since this is not required here and its pointing to my fork.

Like this? https://github.com/ComPWA/tensorwaves/pull/22/commits/b9004ab75c1e3875c31eb6f5bbeddbeb4bb1be9e

tox.ini
spflueger: Hmm true. Forgot about some code syntax checks. I just don't like that some stuff will be in the test env but are not needed. Then test related stuff should go into the tests/requirements.txt?

spflueger:
> > Can you also please remove the AmpliTF submodule, since this is not required here and its pointing to my fork.
> 
> Like this? [b9004ab](https://github.com/ComPWA/tensorwaves/commit/b9004ab75c1e3875c31eb6f5bbeddbeb4bb1be9e)

yep

tox.ini
redeboer: Actually `tox` is not really used at this stage. It currently only serves as a config file for `flake8`. But if it is used, `tox` should serve as a command comparable to travis CI, but locally, so you want to have all dev tools installed in the tox env.

spflueger: Yeah, but ideally you would want to have all the requirements for testing there, and only the ones for testing, not some code formatter (autopep8) etc. Right? I know those other dep are minor. I'm just thinking from an idealistic point of view

redeboer: I understand the reasoning. Way out would be something like `requirements-devtools.txt`
```
autopep8==1.5
flake8==3.7.9
flake8-blind-except
flake8-builtins
flake8-docstrings
flake8-import-order
flake8-rst-docstrings
pep8-naming
pre-commit==2.2.0
pylint==2.4.4
tox==3.14.6
```
so that `tox.ini` can contain
```
[testenv:pep8]
passenv = PYTHONPATH
deps =
    -rrequirements-devtools.txt
commands =
    flake8
```
and then `requirements-dev.txt` would look like
```
-r doc/requirements.txt
-r tests/requirements.txt
-r requirements-devtools.txt
```
but I'm not sure if that's the way to go ^^

Maybe it's better to think about this later on once `tox` gets a more thorough implementation (i.e. if we have proper unit tests https://github.com/ComPWA/tensorwaves/issues/19). For the moment, the `tox.ini` file is really more of a placeholder for the `flake8` config (following [convention](https://flake8.pycqa.org/en/latest/user/configuration.html)) along with an elementary setup for `tox`.

spflueger: Wait, so tox is only used for the code format tests? I thought it would be a generally replacement for setup.py in terms of testing.

Well I'm not entirely sure which dep are doing what now, but for me it should look like this:
* requirements.txt: all dep which are necessary to build and run
* tests/dev.txt: all dep which are required for the tests, that would include flake8, tox... (Im referring to the code format checkers)
* req-dev.txt: all dep needed for developers, which includes autopep8, pre-commit, ....

redeboer: > req-dev.txt: all dep needed for developers, which includes autopep8, pre-commit, ....

But that's the thing :) Do the ... include things like pytest and documentation tools? I would say yes, at least, that's how it is [treated in the documentation so far](https://compwa.github.io/contribute.html), that you just run `pip install -r requirements-dev.txt` and you are ready to do any kind of development. (See also https://github.com/ComPWA/pycompwa/pull/85)

requirements.txt
redeboer: Hm okay, problem was fixing the Python version
https://github.com/ComPWA/tensorwaves/pull/22/commits/a2652e664ee226c0eafafc24fda5b65eb50c4283

tox.ini
spflueger: yes dev includes everything. But I'm talking about the other way around. So let me put it this way
req < test-req < dev-req

redeboer: The structure of three requirements files comes from pycompwa (https://github.com/ComPWA/pycompwa/pull/83). We decided to split them up like this so that travis can run different jobs that do only pytest or only doc building (https://github.com/ComPWA/pycompwa/pull/83/commits/9474e09ec86300f627709425e548c7e7be0e2d2e). So I think a nested structure like **req < test-req < dev-req** this is not a good idea.

redeboer: For me the main aim of this PR is that you have standardised way of creating an environment for (in the first place) running tensorwaves (that needs only the `requirements.txt`) and second, collaborating on development (only additional thing you need to do for that is `pip install -r requirements-dev.txt`).

spflueger: ok lets move this decision to later, when we actually setup ci and tests

Issue #24, opened 2020-04-20
Test formatting in Travis CI job


redeboer:
I understand the reasoning. Way out would be something like `requirements-devtools.txt`
```
autopep8==1.5
flake8==3.7.9
flake8-blind-except
flake8-builtins
flake8-docstrings
flake8-import-order
flake8-rst-docstrings
pep8-naming
pre-commit==2.2.0
pylint==2.4.4
tox==3.14.6
```
so that `tox.ini` can contain
```
[testenv:pep8]
passenv = PYTHONPATH
deps =
    -rrequirements-devtools.txt
commands =
    flake8
```
and then `requirements-dev.txt` would look like
```
-r doc/requirements.txt
-r tests/requirements.txt
-r requirements-devtools.txt
```
but I'm not sure if that's the way to go ^^

Maybe it's better to think about this later on once `tox` gets a more thorough implementation (i.e. if we have proper unit tests https://github.com/ComPWA/tensorwaves/issues/19). For the moment, the `tox.ini` file is really more of a placeholder for the `flake8` config (following [convention](https://flake8.pycqa.org/en/latest/user/configuration.html)) along with an elementary setup for `tox`.

_Originally posted by @redeboer in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjU1Mzk3ODI0OnYy/pull_request_review_threads/more_comments_

redeboer:
See also https://github.com/ComPWA/tensorwaves/pull/22 and https://github.com/ComPWA/tensorwaves/issues/19

redeboer:
@spflueger 
I don't think it's a good idea to test the formatting in a travis job as this is already done through Codacy ([codacy.com/ComPWA/tensorwaves](https://app.codacy.com/gh/ComPWA/tensorwaves/])). The formatting checks that are installed through `requirements-dev.txt` are for local use, particularly at the moment when you commit. This forces the developer to incrementally improve the code quality prior to a PR.

spflueger:
But we definitely need them in the CI. So our decision is between Codacy or Travis.
I guess for now we can stick with Codacy and we can always switch to something else later on.
For me the general way the requirements are currently setup is not ideal.
https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments
I would follow this. So having requirements files for each environment
- ci/test
- prod
- dev

[...]

PR #26, opened 2020-04-21
ci: add travis CI

redeboer:
Closes https://github.com/ComPWA/tensorwaves/issues/19
Closes https://github.com/ComPWA/tensorwaves/issues/24

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,3 +44,18 @@ jobs:
       os: osx
       osx_image: xcode11.2
       language: shell
+
+    - name: "Deploy PyPI package"
+      install: echo "Skipping install stage"
+      script: echo "Skipping script stage"
+      before_script: echo "Skipping before_script stage"
+      deploy:
+        edge: true
+        provider: pypi
+        on:
+          tags: true
+        distributions: sdist
+        skip_existing: false
+        user: __token__
+        password:
+          secure: "HNhEQB0wTyYcAKz6+m/4P3SQH74R69c/A9V+4H/FKwXJDK2kLIFBnetVT7DAWOoG8ombsMDhnMc+bLxsi053Y1dxsEpdQFBzZV28Rmi2JstDhXe9u2diYs8FdtXYwOZQTq0AH2ntEKSnTxEGh5twm84VUcaIBpk0gUHA89nW8CXT/TQ0BrnGLCwQJDNvVdjMrlLckkZde6Vw2oZOhKt3l1FKP7rdkA+/G4g5guK2mPlgBXzDZxhimJvB4hQRr5jlCV3fmnlUQbL0VntaspvDwnwizIneb/2o7vYco0ZiL18uAARjJQ+r130t3gJq0raLCngsgVjrzitldpPRl9W5xSMP+hmTyUNZjQNpr+421uWN+jvD1cZmcl+Fe9s/nk/6Lip/C/2G/mtGrOM9+mt8E+XejgZjF6d9f1O0gSHszHS037hRDQ7ikZLbGEL1kXHEwvku8W50j6fom8Bmwk0yKwnbjDt/DMNzZy6aHaRTEOi0uExkW2OxsWAMPbGxZ8rAAhVbNgEJ1KXFhC0bPioj85Wy2w7KJJVL3eSdz60FZYtPrslfSj3wpaS/Zd+FSLKbmDwTRz8WY/7BaHQvdgTGciLAT+5bxbPtkOuw4sai/1z2aZLwEPpNnbBaMY0GVWJSiIHHy+VpvFHj2t00toBY4ORLTLyJqv5Ucdwcni0Yc6g="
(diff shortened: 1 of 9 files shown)


redeboer:
> Closes https://github.com/ComPWA/tensorwaves/issues/24 (?)

@spflueger 
Not yet sure about this, see https://github.com/ComPWA/tensorwaves/issues/24#issuecomment-617240737

setup.py
```diff
@@ -0,0 +1,40 @@
...
+setuptools.setup(
+    name="tensorwaves",
+    version='0.0-alpha2',
+    long_description=long_description,
+    long_description_content_type="text/markdown",
+    url="https://github.com/ComPWA/tensorwaves",
+    packages=setuptools.find_packages(),
+    classifiers=[
+        "Programming Language :: Python :: 3",
+        "License :: OSI Approved :: MIT License",
```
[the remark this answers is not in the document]
redeboer: Ah yeah https://github.com/ComPWA/tensorwaves/pull/26/commits/7521fe424457cab8a0ef3803966d3ae424520b48

spflueger:
#28 should be addressed with this PR

setup.py
```diff
@@ -0,0 +1,36 @@
...
+
+See:
+https://packaging.python.org/guides/distributing-packages-using-setuptools/
+"""
+
+import setuptools
+
+DATA_FILES = [
+    'particle_list.xml',
+    'particle_list.yaml',
```
spflueger, line 12: This is ok for dev right now. Later on we should def fix this (see #29)

setup.py
```diff
@@ -0,0 +1,36 @@
...
+]
+
+with open("README.md", "r") as fh:
+    long_description = fh.read()
+
+setuptools.setup(
+    name="tensorwaves",
+    version='0.0-alpha2',
```
spflueger, line 20: 0.0-alpha. But the whole semantic versioning https://semver.org/ does not really make much sense for us, since we wont bugfix older version and keep older versions alive as well, right?

.travis.yml
```diff
@@ -0,0 +1,68 @@
...
+      install: echo "Skipping install stage"
+      script: echo "Skipping script stage"
+      before_script: echo "Skipping before_script stage"
+      deploy:
+        edge: true
+        provider: pypi
+        on:
+          tags: true
+        distributions: sdist
+        skip_existing: false
```
spflueger, line 65: why is this false?

redeboer: To overwrite potential existing tags (and keep them in sync with the GitHub tags). Though, a bit of trying out with this tells me this `skip_existing` tag is not functional, because PyPI just bounces back if there is already a file there.

.travis.yml
```diff
@@ -0,0 +1,68 @@
...
+
+branches:
+  only:
+    - master
+    - /\d+\.\d+.*/
+  except:
+    - gh-pages
+
+before_install:
+  - pip3 install --upgrade pip
```
spflueger, line 13: would be be better do create a venv?

setup.py
redeboer: Atm not, but this style is required for and interpreted by PyPI. The alpha2 (instead of alpha1) is because I already uploaded a alpha1.

Anyway, once interfaces stabilise it becomes easy to update the package now:

1. Change the `version` argument of `setuptools.setup` that you see here.
2. Create a PR
3. Squash and approve
4. Create a new tag *from the latest, squashed commit* in the master branch of the form digit.digit(whatever else)
5. Push it to the [ComPWA/tensorwaves](https://github.com/ComPWA/tensorwaves/) repo

Travis will take care of the uploading to PyPI

.travis.yml
redeboer: Not necessary because we now use `langauge: python`

setup.py
spflueger: >Atm not, but this style is required for and interpreted by PyPI. 

I see. But why do packages exist right now. Its completely useless right now. Can you delete that version again

redeboer: Idk, just faithfully followed the issue https://github.com/ComPWA/tensorwaves/issues/19

But yes, versions can be deleted afterwards on PyPI

redeboer: Oh and I also needed alpha1 to be able to create a PyPI token :grimacing:

.travis.yml
spflueger: I guess its only necessary if its uploaded for many python versions and environments.

setup.py
spflueger: I understand you need to create a version for trying it out. But I would just change it back to alpha1 and delete the version on pypi again, since it brings absolutely no benefits (or am I wrong here).
Also currently pypi shows the old info (wrong licence or at least inconsistent).

redeboer: Yes, even after removing PyPI complained the version already existed
https://travis-ci.com/github/ComPWA/pycompwa/jobs/321642688#L3186

So this is why I use 0.0, it's totally for testing and the version will be removed once travis works properly

Issue #18, opened 2020-04-17
Port the interface definitions from ComPWA


spflueger:
Define interfaces for

- [x] Intensity/Function
- [x] Kinematics
- [x] DataSet
- [x] Estimator
- [x] Optimizer
- [x] PhspGenerator

**[Estimated effort: 2h - 1 day]**

spflueger:
From problems that appeared in ComPWA and pycompwa, the `Event` structure should not be ported to tensorwaves. Also the `EventCollection` should NOT be an event based structure but just a `dict` or table like structure.

PR #27, opened 2020-04-21
docs: switch to RTD and improve website

redeboer:
Closes https://github.com/ComPWA/tensorwaves/issues/21

Output rendered at
[pwa.readthedocs.io/projects/tensorwaves/en/docs-rtd](https://pwa.readthedocs.io/projects/tensorwaves/en/docs-rtd/)

PR #30, opened 2020-04-22
feat: add interfaces

redeboer:
Closes https://github.com/ComPWA/tensorwaves/issues/18

Defines the basic interfaces for the interaction of the main modules of
tensorwaves:

* Intensity/Function
* Estimator
* Data
* Optimizer

Note that the interfaces are inspired by ComPWA.
https://github.com/ComPWA/ComPWA

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -6,4 +6,4 @@
 init-import=yes
 
 [MESSAGES CONTROL]
-disable=RST203, RST301, E0601
+disable=RST203, RST301
--- a/.pylintrc
+++ b/.pylintrc
@@ -6,4 +6,4 @@
 init-import=yes
 
 [MESSAGES CONTROL]
-disable=RST203, RST301
+disable=RST203, RST301, E0601
--- a/tensorwaves/interfaces.py
+++ b/tensorwaves/interfaces.py
@@ -1,13 +1,12 @@
-"""Defines top level interfaces of tensorwaves.
-
-"""
+"""Defines top-level interfaces of tensorwaves."""
 
 from abc import ABC, abstractmethod
 
 
 class Function(ABC):
     """
-    Defines the interface of a callable function.
+    Interface of a callable function.
+
     The parameters of the model are separated from the domain variables. This
     follows the mathematical definition, in which a function defines its domain
     and parameters. However specific points in the domain are not relevant.
@@ -16,53 +15,55 @@ class Function(ABC):
     getter and setter (see :func:`~Function.parameters`). The reason for this
     separation is to facilitate the events when parameters have changed.
     """
+
     @abstractmethod
     def __call__(self, dataset: dict) -> list:
         """Evaluate the function.
 
         Args:
-            dataset: a dictionary with domain variable names as keys and a
-            an list or `numpy.array` of  with a
+            dataset: a `dict` with domain variable names as keys.
 
         Return:
-            list or :class:`numpy.array` of values
+            `list` or `numpy.array` of values.
         """
 
     @property
     @abstractmethod
     def parameters(self) -> dict:
-        """
-        """
+        """Get `dict` of parameters."""
 
     @abstractmethod
     def update_parameters(self, new_parameters: dict):
-        """
-        """
+        """Update the collection of parameters."""
 
 
 class Estimator(ABC):
+    """Estimator for discrepancy model and data."""
+
     @abstractmethod
     def __call__(self) -> float:
-        pass
+        """Evaluate discrepancy."""
 
     @abstractmethod
     def gradient(self) -> list:
-        pass
+        """Compute the gradient of the data set."""
 
     @property
     @abstractmethod
     def parameters(self) -> dict:
-        pass
+        """Get `dict` of parameters."""
 
     @abstractmethod
     def update_parameters(self, new_parameters: dict):
-        pass
+        """Update the collection of parameters."""
 
 
 class Kinematics(ABC):
+    """Abstract interface for computation of kinematic variables."""
+
     @abstractmethod
     def convert(self, events: dict) -> dict:
-        pass
+        """Convert a set of momentum tuples (events) to kinematic variables."""
 
     @abstractmethod
     def is_within_phase_space(self, events: dict) -> list:
@@ -71,36 +72,40 @@ class Kinematics(ABC):
     @property
     @abstractmethod
     def phase_space_volume(self) -> float:
-        pass
+        """Compute volume of the phase space."""
 
 
 class Optimizer(ABC):
+    """Optimize a fit model to a data set."""
+
     @abstractmethod
     def optimize(self, estimator: Estimator, initial_parameters: dict) -> dict:
-        pass
+        """Execute optimization."""
 
 
 class UniformRealNumberGenerator(ABC):
+    """Abstract class for generating uniform real numbers."""
+
     @abstractmethod
     def __call__(self, size: int,
                  min_value=0.0, max_value=1.0) -> float or list:
-        '''
-        generate random floats in the range from [min_value,max_value)
-        '''
+        """Generate random floats in the range from [min_value,max_value)."""
 
     @property
     @abstractmethod
     def seed(self) -> dict:
-        pass
+        """Get random seed."""
 
     @seed.setter
     @abstractmethod
     def seed(self, value: float):
-        pass
+        """Set random seed."""
 
 
 class PhaseSpaceGenerator(ABC):
+    """Abstract class for generating phase space samples."""
+
     @abstractmethod
     def generate(self, size: int,
                  random_generator: UniformRealNumberGenerator) -> dict:
-        pass
+        """Generate phase space sample."""
--- a/tensorwaves/interfaces.py
+++ b/tensorwaves/interfaces.py
@@ -1,0 +1,91 @@
+from abc import ABC, abstractmethod
+
+
+class Function(ABC):
+    """
+    Defines the interface of a callable function.
+    The parameters of the model are separated from the domain variables. This
+    follows the mathematical definition, in which a function defines its domain
+    and parameters. However specific points in the domain are not relevant.
+    Hence while the domain variables are the argument of the evaluation
+    (:func:`~Function.__call__`), the parameters are controlled via a getter
+    and setter (see :func:`~Function.parameters`). The reason for this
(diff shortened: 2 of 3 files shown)


tensorwaves/interfaces.py
```diff
@@ -0,0 +1,102 @@
...
+        pass
+
+
+class Kinematics(ABC):
+    @abstractmethod
+    def convert(self, events: dict) -> dict:
+        pass
+
+    @abstractmethod
+    def reduce_to_phase_space(self, events: dict) -> dict:
```
redeboer, line 66: Is this necessary? Or does a `is_within_phase_space` suffice?

tensorwaves/interfaces.py
```diff
@@ -0,0 +1,102 @@
+"""Defines top level interfaces of tensorwaves.
+
+"""
+
+from abc import ABC, abstractmethod
+
+
+class Function(ABC):
```
redeboer, line 8: Sufficient for treating intensities? Or should there be a derived intensity class or so?

tensorwaves/interfaces.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+
+    @parameters.setter
+    @abstractmethod
+    def parameters(self, new_parameters: dict):
+        """
+        """
+
+
+class Estimator(ABC):
```
redeboer, line 44: What is an 'estimator' precisely? I can't formulate it in another way than something like 'estimate of the discrepancy between model and data set'. That suggests there should be some link to both a model and a data set.

tensorwaves/interfaces.py
```diff
@@ -0,0 +1,102 @@
...
+        pass
+
+
+class Optimizer(ABC):
+    @abstractmethod
+    def optimize(self, estimator: Estimator, initial_parameters: dict) -> dict:
+        pass
+
+
+class PhaseSpaceGenerator(ABC):
```
redeboer, line 81: Is a phase space generator independent of a uniform generator?

tensorwaves/interfaces.py
```diff
@@ -0,0 +1,102 @@
...
+    @abstractmethod
+    def generate(self, size: int, random_generator) -> dict:
+        pass
+
+
+class UniformRealNumberGenerator(ABC):
```
redeboer, line 87: Does a real number generator have to be uniform?

tensorwaves/interfaces.py
```diff
@@ -0,0 +1,102 @@
...
+    def reduce_to_phase_space(self, events: dict) -> dict:
+        pass
+
+    @property
+    @abstractmethod
+    def phase_space_volume(self) -> float:
+        pass
+
+
+class Optimizer(ABC):
```
redeboer, line 75: An optimizer has to optimize a model. Should some reference to a model be implemented in this abstract interface?

tensorwaves/interfaces.py
spflueger: I like `is_within_phase_space` better

tensorwaves/interfaces.py
spflueger: yeah `Intensity` would be just a special type of `Function` (regarding the return value)

tensorwaves/interfaces.py
spflueger: Yeah its exactly that. But in an interface should not define any state, that why it is this way.
If data and model would be added to `__call__`  then we static property of the comparison is lost. So I don't know any better way

redeboer: What is the static property of the comparison?

tensorwaves/interfaces.py
spflueger: Yeah, the idea here is that we want to have full reproducability or deterministic behavior. So if you chose a specific phase space generator and seed, the output should always be the same. Since you need the random number generation in the generation of phase space events and in possible hit & miss / accept and reject, it was best to hand a random generator to the phase space generator. That avoids mixups (having two random generators and maintaining their state) and also gives flexibility.

tensorwaves/interfaces.py
spflueger: We only need uniform real number generation. I guess if you need something more special you can also define custom distributions that use that uniform generation as input.

tensorwaves/interfaces.py
spflueger: The optimizer get an Estimator and a set of parameters. He has to tweak the parameters so that the Estimator return value becomes as small as possible

tensorwaves/interfaces.py
redeboer: So the following?
```python
def is_within_phase_space(self, event) -> bool:
    """Check whether event lies within phase space."""
```
Or maybe this?
```python
def is_within_phase_space(self, events: dict) -> list:
    """Check which events lie within phase space."""
```

tensorwaves/interfaces.py
redeboer: Then I suppose this comment resolves this https://github.com/ComPWA/tensorwaves/pull/30/files/a2d859cf1dfd332e6e18f4f97a2bad5faea54911#r413062124

tensorwaves/interfaces.py
```diff
@@ -0,0 +1,102 @@
...
+    @abstractmethod
+    def optimize(self, estimator: Estimator, initial_parameters: dict) -> dict:
+        pass
+
+
+class PhaseSpaceGenerator(ABC):
+    @abstractmethod
+    def generate(self, size: int, random_generator) -> dict:
```
redeboer, line 83: Probably
```python
def generate(self, size: int, random_generator: UniformRealNumberGenerator) -> dict:
```
(note the `UniformRealNumberGenerator`)

tensorwaves/interfaces.py
redeboer: See https://github.com/ComPWA/tensorwaves/pull/30/files/a2d859cf1dfd332e6e18f4f97a2bad5faea54911#r413077761
Or should the `random_generator` in `generate` be a `seed` (`int` or `float`)?

tensorwaves/interfaces.py
spflueger: That the model and the data stay fixed for one instance of an estimator. Only parameters of the model (which are also the parameters of the estimator) are the things that will/can change

tensorwaves/interfaces.py
spflueger: Definitely the second one. Because the first thing we had in the beginning and this event based way of processing did not play together well with the vectorized forms in the rest of the code

tensorwaves/interfaces.py
spflueger: yeah sorry, thats what I had before already. (that got lost in my screw up before)

tensorwaves/interfaces.py
redeboer: https://github.com/ComPWA/tensorwaves/pull/30/commits/d55d6b53324828ea50f359b2dae0870fbdae42ea

tensorwaves/interfaces.py
redeboer: https://github.com/ComPWA/tensorwaves/pull/30/commits/abe46ecb56db3f9abca23211a3cee3eaa9c65bce

tensorwaves/interfaces.py
```diff
@@ -0,0 +1,111 @@
...
+    def optimize(self, estimator: Estimator, initial_parameters: dict) -> dict:
+        """Execute optimization."""
+
+
+class UniformRealNumberGenerator(ABC):
+    """Abstract class for generating uniform real numbers."""
+
+    @abstractmethod
+    def __call__(self, size: int,
+                 min_value=0.0, max_value=1.0) -> float or list:
```
redeboer, line 91: Currently, the types are only as hints to the user and for the API (https://github.com/ComPWA/tensorwaves/pull/27) — there are no checks on these. For better typing, it is better to use the [`typing`](https://docs.python.org/3/library/typing.html) module (from Python 3.5, so should be fine) along with [`mypy`](http://mypy-lang.org/). This can also be added as a test to Travis CI.

spflueger: https://stackoverflow.com/questions/33945261/how-to-specify-multiple-return-types-using-type-hints

Btw you are right, this `or` is not valid. I guess its fine we add this in PR #33

.pylintrc
```diff
@@ -6,4 +6,4 @@
 init-import=yes
 
 [MESSAGES CONTROL]
-disable=RST203, RST301
+disable=RST203, RST301, E0601
```
spflueger, line 9: Why did you need to add E0601 (used-before-assignment). Where does this happen here?

redeboer: In https://github.com/ComPWA/tensorwaves/pull/30/commits/abe46ecb56db3f9abca23211a3cee3eaa9c65bce, `UniformRealNumberGenerator` was defined before `PhaseSpaceGenerator` which generated the `E0610` error. That order changed in https://github.com/ComPWA/tensorwaves/pull/30/commits/7693bca69d6d33a164a4308b03dc567d8ada9cb9, but I did not remove the pylint flag.

Should the E0610 ignore flag be removed?

spflueger: yeah

redeboer: https://github.com/ComPWA/tensorwaves/pull/30/commits/daa36ebc6b29118bc82bfccc68d7f6d9dc4c8a08

tensorwaves/interfaces.py
redeboer: https://github.com/ComPWA/tensorwaves/pull/30/commits/541a23598565c09ff105474d92cab4910bf4a9e2

tox.ini
```diff
@@ -23,4 +23,5 @@ exclude =
     __pycache__
 ignore =
     W503
+    RST304 # Unknown interpreted text role
```
redeboer, line 26: Maybe this ignore should actually not be added. The warning it resulted in may have been correct and would have been picked out more accurately by https://github.com/ComPWA/tensorwaves/pull/27.

spflueger: yeah, better we add a custom defined role instead of ignoring all

redeboer: https://github.com/ComPWA/tensorwaves/pull/30/commits/063b260e45411f101b567406bbe92f07e5ca3bd1

PR #32, opened 2020-04-23
ci: add codecov config

redeboer:
From https://github.com/ComPWA/pycompwa/commit/458ece4efaa81b275934a656dfd5a18402bec457

Closes https://github.com/ComPWA/tensorwaves/issues/19


redeboer:
@spflueger I think this one needs to be merged before the others, otherwise their codecov tests will fail

PR #35, opened 2020-04-24
feat: add AmpliTF as submodule

spflueger:
Note that the submodule currently points to a a fork, because some changes have not been merged into the master. This should be relinked later on.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "AmpliTF"]
+	path = AmpliTF
+	url = https://github.com/apoluekt/AmpliTF.git
```
redeboer, line 3: A Git submodule is not a good solution in Python development, so this should be removed in the future once/if `amplitf` becomes a package.

redeboer: > Note that the submodule currently points to a a fork, because some changes have not been merged into the master. This should be relinked later on.

Is this remark outdated? apoluekt is the upstream, right?

redeboer:
Due to the addition of the submodule, the installation instructions need to be modified
https://github.com/ComPWA/tensorwaves/blob/master/doc/install/get-the-source-code.rst

.gitmodules
spflueger: Yes

spflueger: Actually I'm not sure how `python setup.py install` handles this (non dev mode). Because this should be installed into a separate dir of the venv, in order to be found via `import amplitf`

redeboer: Yeah should be solved soon... better to create an issue for this after this PR is merged (which is still necessary to continue development)
https://github.com/ComPWA/tensorwaves/pull/35#discussion_r414524210

spflueger:
I feel like you are right and we should go the extra mile here.

So 
1. make a PR to `apoluekt/AmpliTF.git` for the changes required to run it
2. make a PR to setup pypi and deployment (using github actions)
3. tag the current version

PR #36, opened 2020-04-24
build: add ampltif as dependency

redeboer:
Closes https://github.com/ComPWA/tensorwaves/pull/35


requirements.txt
```diff
@@ -1,5 +1,7 @@
+amplitf
 numpy
 progress
 pyyaml
+sympy
```
spflueger, line 5: Why is that a requirement? Should be included via amplitf right?

redeboer: I saw it is used in `amplitude.py`

spflueger: Oh that was just there because some functions of amplitf are copied in the current master. Once this is replaced with the correct amplitude building code later on, this requirement will disappear.

spflueger: So you can delete that

redeboer: Ok https://github.com/ComPWA/tensorwaves/pull/36/commits/8b1214cde9166b82db82e7f0f3426d853857fa22

Issue #38, opened 2020-04-24
Implement within phase space boundary check in helicity kinematics


spflueger:
Port functionality of ComPWA `HelicityKinematics` for the `is_within_phase_space` method.

Note: For this the general kinematics information has to be passed to the constructor.

**[Estimated effort: 1 day]**

Issue #15, opened 2020-04-17
Add/port phsp volume integration feature


spflueger:

redeboer:
https://github.com/ComPWA/ComPWA/blob/0bf75f9fe0ec42a1a57730e187fad330a1c44e92/Physics%2FPhspVolume.cpp

Issue #23, opened 2020-04-17
Port data module from pycompwa


spflueger:
The new data module in pycompwa by @redeboer already provides all of the needed functionality.
From that existing data module, only a fraction of the code and functionality is needed though. Here one should also be careful not to port the flaws imposed by the design of ComPWA (`EventCollection`).

Issue #13, opened 2020-04-17
Refactor the expertsystem to decouple the xml structure from the internals


spflueger:

redeboer:
Probably also worth to outsource [`particle.py`](https://github.com/ComPWA/tensorwaves/blob/master/tensorwaves/expertsystem/state/particle.py) to the [`Particle` package](https://pypi.org/project/Particle/). It may need some wrapping, but the important part is that its functionality should be usable by the `physics` module as well.

See also https://github.com/ComPWA/tensorwaves/pull/39#discussion_r415830812

PR #39, opened 2020-04-24
feat: implement basic kinematics functionality of the helicity formalism

### LINKED ISSUE #3: feat: add kinematics functionality
Port the ComPWA kinematics functionality to tensorwaves. The helicity kinematics is responsible for converting event based data into a dataset which can be used to evaluate an Intensity.

**https://github.com/ComPWA/tensorwaves/issues/18 (interfaces) should be closed first**

spflueger:
Closes #3 

Note: Some functionality of the ComPWA kinematics is still missing #15, #38.


redeboer:
[Failed job](https://travis-ci.com/github/ComPWA/tensorwaves/jobs/323090615) will be fixed by #42

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -1,16 +1,262 @@
...
+    def recoil_state(self):
+        """Get final state content of the recoil partner."""
+        return self._recoil_state
+
+    @property
+    def parent_recoil_state(self):
+        """Get final state content of the recoil partner of the parent."""
+        return self._parent_recoil_state
+
+    def __eq__(self, other):
```
redeboer, line 66: Currently, only the special methods `__call__` and `__init__` are rendered in the API.

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -1,16 +1,262 @@
...
+        state_fs = []
+        for fs_uid in subsystem.final_states:
+            state_fs += fs_uid
+        invmass_name = self.register_invariant_mass(list(set(state_fs)))
+        angle_names = self.register_helicity_angles(subsystem)
+
+        return (invmass_name,) + angle_names
+
+    def _convert_ids_to_indices(self, ids: tuple):
+        """Convert unique ids to event indices.
```
redeboer, line 193: About the docstring here: private methods are currently not rendered in the API. Add them as well?

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -1,16 +1,262 @@
...
+from typing import Union
+
+import amplitf.kinematics as tfa_kin
+
+import numpy as np
+
+from tensorwaves.interfaces import Kinematics
+
+
+class SubSystem():
```
redeboer, line 25: How specific is the `SubSystem` to the helicity formalism? If not, it may be an idea to put it into the `physics` module itself.

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -1,16 +1,262 @@
+r"""Kinematic based calculations for the helicity formalism.
```
redeboer, line 1: We may have to discuss the file structure further. For instance, if other formalisms are implemented, something like this may be more convenient:
```python
from tensorwaves.physics.formalism.helicity import Kinematics
```
(In that case, though `class HelicityKinematics` would have to be defined as `class Kinematics(interfaces.Kinematics)`.)

Something similar applies to for instance `PhaseSpaceGenerator`, I mean, that you can just replace their implementation by modifying the imports without for instances having to replace all instances of `HelicityFormalism` in your script.

But I'm not sure how conventional such a structure where we rely more on namespaces would be. The motivation is initially more the file structure, but also that you normally don't mix formalisms or generators (although not even sure about that).

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: I would def not render them, as they are not part of the API (just some internal function). Maybe we should disable this kind of docstring in the flake8/pylint settings as well. Maybe these docstrings are helpful for developers though.

redeboer: Yeah and, in addition, they are still available through `help`.

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Good point. Should we move it to `physics/__init__.py`?

redeboer: Depends, see https://github.com/ComPWA/tensorwaves/pull/39/files#r415805711
In that case it could also be part of `formalism`. Not sure, depends where else a `SubSystem` is used.

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Hmm, I think we should add this since its part of the public API. A user can do `SubSystem1 == SubSystem2`

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -1,16 +1,262 @@
...
+    * :attr:`final_states`
+    * :attr:`recoil_state`
+    * :attr:`parent_recoil_state`
+    """
+
+    def __init__(self, final_states, recoil_state, parent_recoil_state):
+        """Fully initializes the :class:`~SubSystem`.
+
+        Args:
+            final_states: `tuple` of `tuple` s containing unique ids.
```
redeboer, line 40: The unique IDs come from ComPWA, but I wonder if there is not an other way to do this, for instance with the original particle names as strings (or their PIDs), but an index added to duplicate entries. In a way also related to the discusion about `EventCollection` (https://github.com/ComPWA/tensorwaves/issues/23).

redeboer: I mention it also because I see problems coming up with labels like these again
https://github.com/ComPWA/tensorwaves/pull/39/files#diff-1d2eb2f144ec58acdde4c21d96eede1bR138

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -1,16 +1,262 @@
...
+            A `str` key representing the invariant mass. It can be used to
+            retrieve this invariant mass from the dataset returned by
+            :meth:`~convert`.
+        """
+        logging.debug("registering inv mass in kinematics")
+        final_state = tuple(final_state)
+        if final_state not in self._registered_inv_masses:
+            label = 'mSq'
+            for particle_uid in final_state:
+                label += '_' + str(particle_uid)
```
redeboer, line 139: See https://github.com/ComPWA/tensorwaves/pull/39/files#r415812813

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -1,16 +1,262 @@
...
+            They can be used to retrieve the kinematic data from the dataset
+            returned by :meth:`~convert`.
+        """
+        state_fs = []
+        for fs_uid in subsystem.final_states:
+            state_fs += fs_uid
+        invmass_name = self.register_invariant_mass(list(set(state_fs)))
+        angle_names = self.register_helicity_angles(subsystem)
+
+        return (invmass_name,) + angle_names
```
redeboer, line 190: Does this agree with the description of the return value? What does `(invmass_name,)` do?

redeboer:
Good work, also nicely documented! If this module is required to move on with the development, I'm happy to approve and discuss the bigger issues (final state ID and namespaces) separately.

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: yes, it's just concatenating two tuples

redeboer: Ah nvm I misread sorry

tensorwaves/physics/helicityformalism/kinematics.py
redeboer: It can be added here
https://github.com/ComPWA/tensorwaves/blob/4a06430ab9d070e0c879132c212cf42f854dadef/doc/conf.py#L62

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Good point. Actually since its python now, there is actually no type defined on this unique id. It just has to uniquely identify a particle from the event collection later on.
Regarding this fact and that this would also require to change the model recipe file description (refer to names instead of ints), this deserves a separate issue and discussion.

tensorwaves/physics/helicityformalism/kinematics.py
redeboer: Let's leave it here for now an worry about it once an other formalism is implemented.

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Lets move this to a separate discussion. Can you open an issue for that and link it here.

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Ok I will add that

tensorwaves/physics/helicityformalism/kinematics.py
redeboer: Hm perhaps better not to pursue this indeed:
> Flat is better than nested.
https://inventwithpython.com/blog/2018/08/17/the-zen-of-python-explained/

redeboer: But I do think it's better to use `tensorwaves.physics.formalism.helicity` instead of `tensorwaves.physics.helicityformalism`. Or maybe `formalisms`. Then `physics` has at least `dynamics` and `formalisms`.

tensorwaves/physics/helicityformalism/kinematics.py
redeboer: Maybe better to decide on this while refactoring the expert system? https://github.com/ComPWA/tensorwaves/issues/13

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: See https://github.com/ComPWA/tensorwaves/pull/39/commits/8cb12e031798e40cbfb6a3a094e8dca7f91c3ce4

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: I think the dynamics is also strongly coupled to the formalism

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Yeah, and like I tried to say above, this module is actually not limiting the functionality here. Its more the way events are generated and what mapping is handed to the kinematics instance. Btw you should resolve this if you are happy with it

redeboer: Yeah sure, it's just that the question mark left it open

tensorwaves/physics/helicityformalism/kinematics.py
redeboer: Okay leave it like this

Issue #33, opened 2020-04-23
Add static typing checks (mypy) and test in Travis CI job


redeboer:
Currently, the types are only as hints to the user and for the API (https://github.com/ComPWA/tensorwaves/pull/27) — there are no checks on these. For better typing, it is better to use the [`typing`](https://docs.python.org/3/library/typing.html) module (from Python 3.5, so should be fine) along with [`mypy`](http://mypy-lang.org/). This can also be added as a test to Travis CI.

_Originally posted by @redeboer in https://github.com/ComPWA/tensorwaves/pull/30/review_comment/create_

- [ ] Add mypy to requirements and pre-commit https://github.com/ComPWA/tensorwaves/pull/48
- [ ] Fix mypy errors in non-expertsystem modules
- [ ] Add travis check on mypy https://github.com/ComPWA/tensorwaves/pull/48/commits/1f81304f80a47309f9082535075ca5ae08d10bdc

redeboer:
In addition: [Codacy](https://app.codacy.com/gh/ComPWA/tensorwaves/) is nice from a graphical point of view, but the linting checks there do not fully agree with the checks of pre-commit. For instance, all flake checks there are not performed.

On the other hand, if you would perform linting checks through Travis, there is no option anymore to temporarily ignore certain checks (as there is on Codacy).

PR #47, opened 2020-04-28
docs!: change documentation policy

redeboer:
This commit remove the need to write redundant docstrings everywhere.

- [x] Render type hints as parameters and return type in Sphinx documentation
- [x] Remove `__init__` from `special-members` (i.e. merge documentation with class entry)
- [x] Remove `pylint` checks on method docstrings --> already added `C0116`, see [this discussion](https://stackoverflow.com/questions/7877522/how-do-i-disable-missing-docstring-warnings-at-a-file-level-in-pylint)
- [x] Remove `flake8` checks on function docstrings.

PR #48, opened 2020-04-28
ci: add mypy configuration and pre-commit hook

redeboer:
Closes https://github.com/ComPWA/tensorwaves/issues/33

See also https://github.com/ComPWA/tensorwaves/pull/47

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ dist/
 .coverage.*
 .eggs
 .ipynb_checkpoints
+.mypy*/
 .pytest_cache/
 __pycache__
 htmlcov/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,12 @@ repos:
         language: system
         types:
           - python
+      - id: mypy
+        name: mypy
+        entry: mypy --config-file=tox.ini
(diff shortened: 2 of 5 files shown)


redeboer:
Draft for now because we need to see what configurations to add
https://mypy.readthedocs.io/en/stable/config_file.html

I'm not sure if we should add yet another config file, in this case `mypy.ini`. So I would prefer to add the `mypy` configuration to the `tox.ini` file. See this discussion though... https://github.com/python/mypy/pull/2761.

redeboer:
Indeed. An attempt to improve the contribute pages was drafted in https://github.com/ComPWA/pycompwa/pull/85, but that PR should probably be migrated to tensorwaves. Then the instructions can be simplified, we now only use:

* formatter: `black`
* linting: `pylint` + `flake8` (+ extensions)
* static type checker: `mypy`

All of that is automatised through pre-commit and Travis CI, just need to point out those four tools and their corresponding configs. The configs are important because they represent our (if necessary, evolving) coding conventions, so those conventions don't need to be written out in the documentation.

Related to that, as you say, we may want to highly recommend contributors to work with VS code. There is currently no such assumption in the documentation.

PR #49, opened 2020-04-28
style: add black formatting configuration

redeboer:
This is a proposal presented as a PR (instead of an issue) as to be able to directly see the effect. The formatting is also tested in Travis CI.

See also https://github.com/ComPWA/pycompwa/pull/91

### DIFF
--- a/doc/contribute.rst
+++ b/doc/contribute.rst
@@ -7,9 +7,6 @@ Some recommended packages for Python development
 * `pytest <https://docs.pytest.org/en/latest/>`_: Run :code:`pytest` in the main
   folder of the repository to run all :file:`test_*.py` files
 
-* `autopep8 <https://pypi.org/project/autopep8/0.8/>`_: Auto-format your Python
-  code
-
 * `pylint <https://www.pylint.org/>`_: Scan your code for naming conventions
   and proper use of Python
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,27 @@
+[build-system]
+requires = [
+    "setuptools",
+    "wheel",
+]
+
+[tool.black]
+line-length = 79
+target-version = [
+    'py36',
+    'py37',
+]
+include = '\.pyi?$'
+exclude = '''
+/(
+    .*\.egg-info
+  | .*build
+  | \.ComPWA
+  | \.eggs
+  | \.git
+  | \.pytest_cache
+  | \.tox
+  | \.venv
+  | \.vscode
+  | dist
+)/
+'''
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ exclude = '''
 /(
     .*\.egg-info
   | .*build
-  | \.ComPWA
   | \.eggs
   | \.git
   | \.pytest_cache
--- a/tox.ini
+++ b/tox.ini
@@ -22,14 +22,14 @@ filename =
 exclude =
     __pycache__
     doc/conf.py
-ignore =
+ignore = # more info: https://www.flake8rules.com/
     D102 # method docstring
     D103 # function docstring
     D107 # init docstring
-    E203 # black
-    E501 # black
+    E203 # https://github.com/psf/black#slices
+    E501 # https://github.com/psf/black#line-length
     RST301 # unexpected indentation (related to google style docstring)
-    W503
+    W503 # https://github.com/psf/black#line-breaks--binary-operators
 application-import-names = tensorwaves
 rst-roles =
     class,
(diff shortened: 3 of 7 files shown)


pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[build-system]
+requires = [
+    "setuptools",
+    "wheel",
+]
+
+[tool.black]
+line-length = 79
```
spflueger, line 8: why not 80?

spflueger: Nevermind. Strange I thought pep8 would also be 80 chars

pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+target-version = [
+    'py36',
+    'py37',
+]
+include = '\.pyi?$'
+exclude = '''
+/(
+    .*\.egg-info
+  | .*build
+  | \.ComPWA
```
spflueger, line 18: I guess thats a copy/paste error

pyproject.toml
redeboer: > The limits are chosen to avoid wrapping in editors with the window width set to 80
https://www.python.org/dev/peps/pep-0008/#maximum-line-length
Also read somewhere that it is for files where the newline character is not counted

pyproject.toml
redeboer: Oops yes... https://github.com/ComPWA/pycompwa/pull/91

redeboer: https://github.com/ComPWA/tensorwaves/pull/49/commits/bb9825165864f9da1a6de14972c3329cf30a1e09

tox.ini
```diff
@@ -26,6 +26,9 @@ ignore =
     D102 # method docstring
     D103 # function docstring
     D107 # init docstring
+    E203 # black
+    E231 # black
+    E501 # black
```
spflueger, line 31: I feel like we should add a bit more clear comment why these are inserted (non pep8 compliant)

doc/contribute.rst
```diff
@@ -7,9 +7,6 @@ Some recommended packages for Python development
 * `pytest <https://docs.pytest.org/en/latest/>`_: Run :code:`pytest` in the main
   folder of the repository to run all :file:`test_*.py` files
 
-* `autopep8 <https://pypi.org/project/autopep8/0.8/>`_: Auto-format your Python
-  code
-
```
spflueger, line 12: Do you want to mention black here?

pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+line-length = 79
+target-version = [
+    'py36',
+    'py37',
+]
+include = '\.pyi?$'
+exclude = '''
+/(
+    .*\.egg-info
+  | .*build
```
spflueger, line 17: Is build also necessary? I'm not sure setuptools will create something like this?

doc/contribute.rst
redeboer: Should be, but the contribute page is about to be rewritten anyway https://github.com/ComPWA/pycompwa/pull/85

pyproject.toml
spflueger: Hmm ok sphinx will create a build dir. But I have to read up on how to read this syntax here

redeboer: I'm not sure, it came from black's [pyproject.toml example](https://black.readthedocs.io/en/stable/pyproject_toml.html) originally, then left it in for skbuild in pycompwa. I don't think it hurts to leave it in.

redeboer: > I have to read up on how to read this syntax here

It's regex (only not sure about the `|`).

spflueger: black default for exclude is:
[default: /(\.eggs|\.git|\.hg|\.mypy_cache|\.nox|\.tox|\.venv|_build|buck-out|build|dist)/]

Not sure why they did not make a `.*` for build

doc/contribute.rst
spflueger: ok

pyproject.toml
spflueger: > It's regex (only not sure about the |).

I assume its logical or

spflueger: I guess its fine. I can't think of any reason why `.*build` should cause problems. We can also change it later on if it does

tox.ini
redeboer: https://github.com/ComPWA/tensorwaves/pull/49/commits/62e5d1e044829be755c55de3ddce81475dde5fdf

pyproject.toml
redeboer: I added the regex `.*` for 'anything preceding it'

Issue #44, opened 2020-04-27
Refactor amplitude building code


spflueger:
The current code is combining building and computation/execution which is inconvenient for testing.

So as an example rather than
```python
class _CoefficientAmplitude():
    def __init__(self, recipe: dict, builder):
        if isinstance(recipe, dict):
            pars = recipe['Parameters']
            self.mag = builder._get_parameter(pars['Magnitude'])
            self.phase = builder._get_parameter(pars['Phase'])

            self.amp = builder._create_amplitude(recipe['Amplitude'])

    def __call__(self, x):
        coefficient = atfi.complex(
            self.mag * atfi.cos(self.phase),
            self.mag * atfi.sin(self.phase)
        )
        return coefficient * self.amp(x)
```

define an computation part
```python
class _CoefficientAmplitude():
    def __init__(self, amplitude, mag, phase):
            self.mag = mag
            self.phase = phase
            self.amp = amplitude

    def __call__(self, x):
        coefficient = atfi.complex(
            self.mag * atfi.cos(self.phase),
            self.mag * atfi.sin(self.phase)
        )
        return coefficient * self.amp(x)
```
and a building part
```python
def create_cofficient_amplitude(self, recipe: dict, builder):
    if isinstance(recipe, dict):
        pars = recipe['Parameters']
        mag = builder._get_parameter(pars['Magnitude'])
        phase = builder._get_parameter(pars['Phase'])
        amp = builder._create_amplitude(recipe['Amplitude'])
        return _CoefficientAmplitude(amp, mag, phase)
```

spflueger:
I propose to make these changes after the current code is uploaded so more concurrent work can go on

PR #56, opened 2020-04-30
refactor: improve amplitude building code

### LINKED ISSUE #2: feat: add parameter handling for amplitude builder

spflueger:
Closes #2

This will be improved further with #44

### DIFF
--- a/tensorwaves/physics/helicityformalism/amplitude.py
+++ b/tensorwaves/physics/helicityformalism/amplitude.py
@@ -1,9 +1,9 @@
 """Amplitude module for the helicity formalism.
 
 Its responsibility is the construction of complicated helicity formalism
-amplitude models using the a recipe (see `IntensityBuilder`). These models
-are encapsulated in an `IntensityTF` class, which can be evaluated as a
-regular callable.
+amplitude models using a recipe (see `IntensityBuilder`). These models are
+encapsulated in an `IntensityTF` class, which can be evaluated as a regular
+callable.
 """
 
 import logging
@@ -41,7 +41,7 @@ class IntensityTF(Function):
         """Evaluate the Intensity.
 
         Args:
-            dataset: A dict of kinematic variables.
+            dataset: Contains all required kinematic variables.
 
         Returns:
             List of intensity values.
@@ -64,10 +64,12 @@ class IntensityBuilder:
     """Builds Intensities from helicity formalism recipe files.
 
     Args:
-        particles: A dictionary containing info of various particles.
-        kinematics: A helicity kinematics instance.
-        phsp_data: An optional phase space event collection, which is required
-            if a normalization of the Intensity is performed.
+        particles: Contains info of various particles.
+        kinematics: A helicity kinematics instance. Note that this kinematics
+            instance will be modified in the process.
+        phsp_data: A phase space event collection, required if a normalization
+            of the Intensity is performed.
+
     """
 
     def __init__(
@@ -86,8 +88,8 @@ class IntensityBuilder:
         """Create an `IntensityTF` instance based on a recipe.
 
         Args:
-            recipe: A dict containing builder instructions. These recipe
-                files can be generated via the expert system
+            recipe: Contains builder instructions. These recipe files can be
+                generated via the expert system
                 (see `~.expertsystem.amplitude.helicitydecay.HelicityAmplitudeGeneratorXML`)
 
         """
@@ -123,7 +125,7 @@ class IntensityBuilder:
                 recipe, self, self._kinematics
             )
 
-        raise Exception("Unknown intensity {}!".format(intensity_class))
+        raise Exception(f"Unknown intensity {intensity_class}!")
 
     def _create_amplitude(self, recipe: dict) -> Callable:
         amplitude_class = recipe["Class"]
@@ -144,7 +146,7 @@ class IntensityBuilder:
                 recipe, self, self._kinematics, self._particles, self._dynamics
             )
 
-        raise Exception("Unknown amplitude {}!".format(amplitude_class))
+        raise Exception(f"Unknown amplitude {amplitude_class}!")
 
     def _register_parameter(self, name: str, value: float) -> tf.Variable:
         if name not in self._parameters:
@@ -156,8 +158,8 @@ class IntensityBuilder:
     def _get_parameter(self, name: str) -> tf.Variable:
         if name not in self._parameters:
             raise Exception(
-                "Parameter {} not registered! Your recipe file is"
-                " corrupted!".format(name)
+                "Parameter {name} not registered! Your recipe file is"
+                " corrupted!"
             )
 
         return self._parameters[name]
@@ -169,6 +171,11 @@ class IntensityBuilder:
             )
 
     def _get_normalization_data(self) -> Optional[dict]:
+        if not self._phsp_data:
+            raise Exception(
+                "No phase space sample given! This is required for the "
+                "normalization."
+            )
         return self._phsp_data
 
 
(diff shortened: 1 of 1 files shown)


spflueger:
Ok, the PR is rebased to the master (which includes PR #58)

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -1,246 +1,402 @@
...
+    def create_intensity(self, recipe: dict) -> IntensityTF:
+        """Create an `IntensityTF` instance based on a recipe.
+
+        Args:
+            recipe: A dict containing builder instructions. These recipe
+                files can be generated via the expert system
+                (see `~.expertsystem.amplitude.helicitydecay.HelicityAmplitudeGeneratorXML`)
+
+        """
+        self._dynamics = recipe["Dynamics"]
```
redeboer, line 94: We may have to create some sort of object representing a recipe structure, because it's hard to debug string keys. In general it is not advised to work with a nested structure of dictionaries for a specific data structure. In addition, it assumes to a large degree we are working with YAML recipes here.

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -1,246 +1,402 @@
...
+        if name not in self._parameters:
+            self._parameters[name] = tf.Variable(
+                value, name=name, dtype=tf.float64
+            )
+        return self._parameters[name]
+
+    def _get_parameter(self, name: str) -> tf.Variable:
+        if name not in self._parameters:
+            raise Exception(
+                "Parameter {} not registered! Your recipe file is"
```
redeboer, line 159: Since we have given up support, we can make use of [literal string interpolation](https://www.python.org/dev/peps/pep-0498/), which is nicer to read.

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -1,246 +1,402 @@
...
 
-class _CoherentIntensity(tf.keras.Model):
-    def __init__(self, recipe: dict):
-        super(_CoherentIntensity, self).__init__(name="CoherentIntensity")
+class _CoherentIntensity:
+    def __init__(self, recipe: dict, builder):
         if isinstance(recipe["Amplitudes"], list):
-            amp_recipes = [v for v in recipe["Amplitudes"]]
-            self.amps = [_create_amplitude(x) for x in amp_recipes]
+            amp_recipes = list(recipe["Amplitudes"])
```
redeboer, line 222: Is this conversion to `list` necessary?

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -1,246 +1,402 @@
...
+            ]
+            if not decay_dynamics:
+                raise LookupError(
+                    "Could not find dynamics for particle with "
+                    "name " + decaying_state["Name"]
+                )
+            decay_dynamics = decay_dynamics[0]
+
+            non_dynamics_label = "nonDynamic"
+            known_dynamics_functions = {
```
redeboer, line 329: I wonder if we should move these kind of expandable dicts (implementations) to a move visible place, because I'm not sure how many steps need to be taken in order to implement a new dynamic function with this set-up.

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -1,246 +1,402 @@
...
+                "relativisticBreitWigner": relativistic_breit_wigner,
+            }
+
+            if non_dynamics_label in decay_dynamics["Type"]:
+                self._call_wrapper = self._without_dynamics
+            else:
+                if decay_dynamics["Type"] not in known_dynamics_functions:
+                    raise ValueError(
+                        "Dynamics ("
```
redeboer, line 338: See remark at https://github.com/ComPWA/tensorwaves/pull/56/files#diff-f1d492f82016d212f95e0d4e57fb0651R159

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -1,246 +1,402 @@
...
-        "HelicityDecay": _HelicityDecay,
-    }
-    if amplitude_class in known_amplitude_builders:
-        return known_amplitude_builders[amplitude_class](recipe)
-    else:
-        logging.error(
-            "unknown intensity " + str(amplitude_class) + "! Skipping"
+        raise Exception("Unknown intensity {}!".format(intensity_class))
+
+    def _create_amplitude(self, recipe: dict) -> Callable:
```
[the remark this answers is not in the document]
spflueger: I try to make only things public if the user should have access. For only the builder needs the access. Also keep in mind #44.

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: Yeah, I wanted to address this later for example in #44 .

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -1,246 +1,402 @@
...
+    def __init__(self, tf_model: Callable, parameters: Dict[str, tf.Variable]):
+        self._model = tf_model
+        self.__parameters = parameters
 
-# copied from TensorFlowAnalysis
+    def __call__(self, dataset: Dict[str, numpy.ndarray]) -> numpy.ndarray:
+        """Evaluate the Intensity.
 
+        Args:
+            dataset: A dict of kinematic variables.
```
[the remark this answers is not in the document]
spflueger: The sphinx autodoc from the types will automatically insert the correct type. So I will just change the wording a bit to get rid of the dict

tensorwaves/physics/helicityformalism/amplitude.py
redeboer: > On the top of my head I only see that you have no "tab completion" for the contents if you use strings.

It's also the fact that you will find typos in the strings only at runtime.

> How so? We only use a dict of strings and basic python types.

I remember mainly the mess of using XML, where there would be tags and attributes, some with an `@` prepended.

spflueger: > It's also the fact that you will find typos in the strings only at runtime.

Yeah, thats what I mean with "tab completion" :D. But to test the functionality you also have to write unit tests, right? And the whole idea is that the recipe is dynamic. So I will only know at runtime (once I pass it to the builder) what is actually inside this. Maybe we should move this discussion to a separate issue

>I remember mainly the mess of using XML, where there would be tags and attributes, some with an @ prepended.

Yeah that was xml language specific (attribute). Now I just require a nested dictionary of those special keys. So it should also work fine with the xml, if the attributes are changed into tags.

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: https://github.com/ComPWA/tensorwaves/pull/56/commits/5f7937d96cda5d51460ec1900d0d84e13d2ddf24

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: https://github.com/ComPWA/tensorwaves/pull/56/commits/5f7937d96cda5d51460ec1900d0d84e13d2ddf24

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: https://github.com/ComPWA/tensorwaves/pull/56/commits/5f7937d96cda5d51460ec1900d0d84e13d2ddf24

Issue #59, opened 2020-05-05
Deterministic behavior for phasespace generator


spflueger:
The phase space generator using the `phasespace` package has to be improved to accept a random generator. This will help to ensure deterministic behavior based on the seed. Or at least the deterministic behavior has to be verified.

So TFPhaseSpaceGenerator and TFUniformRealNumberGenerator are currently not decoupled, because the `phasespace` package uses `tf.random` internally as well. This is currently in conflict with our generate definition and can be problematic. The random generators have to be kept in sync. We might have to think about our generate definition as well.

**[Estimated effort: 1-5 days]**

redeboer:
The `phasespace` package makes uses of [`tf.random.uniform`](https://www.tensorflow.org/api_docs/python/tf/random/uniform)
https://github.com/zfit/phasespace/blob/6e442c7644482120c0efe42f0a85ed46890da364/phasespace/phasespace.py#L326

This is an old RNG that TensorFlow  [discourages from TF1.x on](https://www.tensorflow.org/guide/random_numbers). Instead they recommend using [`tf.random.Generator`](https://www.tensorflow.org/api_docs/python/tf/random/Generator).

mayou36:
This is indeed outdated and should be updated. I am not quite sure myself for the best strategy of this however.

I think the most pragmatic solution is to have a global generator that can be set (for any package). This would fulfill the two requirements we have (usually):
- random numbers are random and not correlated (just recently ran into this when doing convolution by addition...). Since one global generator is used, this is fulfilled.
- deterministic behavior: same code with same generator will result in the same result. However, a single model won't output the same with the same seed (since it's a global seed), but then this is usually not required.

Why I would not have a "per instance" generator or something: I don't see a real usecase for this and it just adds to the complexity.

spflueger:
Yes, we need a single source of truth for the random numbers to guarantee the two points you mentioned.

The two possible options are:

1. Global generator as mentioned above
2. Feeding a single random number generator instance to the event generation facade code. The facade code will ensure that only that one rng instance is used.

I prefer option 2, as it is more flexible and modular. Furthermore I believe option 2 is actually also easier to implement, since all that has to be done in the facade is hand the rng instance to the phase space event generator. The only tricky part might be tensorflow itself, that injecting a regular python callable (interface to rng) requires adjustments. But it would be fine if the phasespace package would just allow the tensorflow random generator as an option.

What do you think @redeboer , @mayou36 ?

[...]

PR #60, opened 2020-05-05
feat: add data generation functionality

spflueger:
We might have to improve the generate interface slightly. This is currently ported 1 to 1 from ComPWA.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,13 +4,13 @@ codecov:
 coverage:
   precision: 2
   round: down
-  range: "75...100"
+  range: "60...90" # color range
 
   status:
     project:
       default:
         # basic
-        target: 75% # can't go below 75%
+        target: 60% # can't go below 75%
         threshold: 3% # allow drops by 3%
         base: auto
         # advanced
--- a/examples/intensity-recipe.yaml
+++ b/examples/intensity-recipe.yaml
@@ -1169,7 +1169,7 @@ Intensity:
                     RecoilFinalState: 2
 
 ParticleList:
-  - Name: J/psi
+  J/psi:
     PID: 443
     Mass: { Value: 3.096900 }
     Width: { Value: 9.29E-05 }
@@ -1184,7 +1184,7 @@ ParticleList:
       Charm: 0
       Strangeness: 0
 
-  - Name: gamma
+  gamma:
     PID: 22
     Mass: { Value: 0.0 }
     QuantumNumbers:
@@ -1197,7 +1197,7 @@ ParticleList:
       Charm: 0
       Strangeness: 0
 
-  - Name: pi0
+  pi0:
     PID: 111
     Mass: { Value: 0.1349766, Error: 0.000006 }
     QuantumNumbers:
@@ -1211,7 +1211,7 @@ ParticleList:
       Charm: 0
       Strangeness: 0
 
-  - Name: f0(980)
+  f0(980):
     PID: 9010221
     Mass: { Value: 0.994, Error: 0.001 }
     Width: { Value: 0.070 }
@@ -1223,7 +1223,7 @@ ParticleList:
       IsoSpin: { Value: 0 }
       Gparity: 1
 
-  - Name: f2(1270)
+  f2(1270):
     PID: 225
     Mass: { Value: 1.2751, Error: 0.0012 }
     Width: { Value: 0.185 }
@@ -1234,7 +1234,7 @@ ParticleList:
       Cparity: 1
       Gparity: 1
 
-  - Name: f0(1500)
+  f0(1500):
     PID: 9030221
     Mass: { Value: 1.505, Error: 0.006 }
     Width: { Value: 0.109 }
@@ -1245,7 +1245,7 @@ ParticleList:
       Cparity: 1
       Gparity: 1
 
-  - Name: f2(1950)
+  f2(1950):
     PID: 9050225
     Mass: { Value: 1.944, Error: 0.012 }
     Width: { Value: 0.472 }
--- a/examples/intensity-recipe.yaml
+++ b/examples/intensity-recipe.yaml
@@ -12,15 +12,13 @@ Kinematics:
       ID: 4
 
 Parameters:
-  - Name: &par0 strength_incoherent
-    Type: Strength
-    Value: 1
-    Fix: true
   - Name: &par1 Magnitude_J/psi_to_f2(1270)_0+gamma_-1;f2(1270)_to_pi0_0+pi0_0;
     Type: Magnitude
     Value: 1.0
+    Fix: true
   - Name: &par2 Phase_J/psi_to_f2(1270)_0+gamma_-1;f2(1270)_to_pi0_0+pi0_0;
     Type: Phase
+    Value: 0.0
   - Name: &par3 Magnitude_J/psi_to_f2(1950)_0+gamma_-1;f2(1950)_to_pi0_0+pi0_0;
     Type: Magnitude
     Value: 1.0
@@ -63,17 +61,9 @@ Parameters:
   - Name: &par16 Phase_J/psi_to_f0(1500)_0+gamma_-1;f0(1500)_to_pi0_0+pi0_0;
     Type: Phase
     Value: 0.0
-  - Name: &par17 Magnitude_J/psi_to_omega(782)_1+pi0_0;omega(782)_to_gamma_-1+pi0_0;
-    Type: Magnitude
-    Value: 1.0
-  - Name: &par18 Phase_J/psi_to_omega(782)_1+pi0_0;omega(782)_to_gamma_-1+pi0_0;
-    Type: Phase
-    Value: 0.0
 
 Intensity:
-  Class: StrengthIntensity
-  Component: incoherent_with_strength
-  Strength: *par0
+  Class: NormalizedIntensity
   Intensity:
     Class: IncoherentIntensity
     Intensities:
@@ -82,7 +72,9 @@ Intensity:
         Amplitudes:
           - Class: CoefficientAmplitude
             Component: J/psi_-1_to_f2(1270)_0+gamma_-1;f2(1270)_0_to_pi0_0+pi0_0;
-            Parameters: [*par1, *par2]
+            Parameters:
+              Magnitude: *par1
+              Phase: *par2
             Amplitude:
               Class: SequentialAmplitude
               Amplitudes:
@@ -114,7 +106,9 @@ Intensity:
                     RecoilFinalState: 2
           - Class: CoefficientAmplitude
             Component: J/psi_-1_to_f2(1950)_0+gamma_-1;f2(1950)_0_to_pi0_0+pi0_0;
-            Parameters: [*par3, *par4]
+            Parameters:
+              Magnitude: *par3
+              Phase: *par4
             Amplitude:
               Class: SequentialAmplitude
               Amplitudes:
@@ -146,7 +140,9 @@ Intensity:
                     RecoilFinalState: 2
           - Class: CoefficientAmplitude
             Component: J/psi_-1_to_f2(1270)_-1+gamma_-1;f2(1270)_-1_to_pi0_0+pi0_0;
-            Parameters: [*par5, *par6]
+            Parameters:
+              Magnitude: *par5
+              Phase: *par6
             Amplitude:
               Class: SequentialAmplitude
               Amplitudes:
@@ -178,7 +174,9 @@ Intensity:
                     RecoilFinalState: 2
           - Class: CoefficientAmplitude
             Component: J/psi_-1_to_f2(1950)_-1+gamma_-1;f2(1950)_-1_to_pi0_0+pi0_0;
-            Parameters: [*par7, *par8]
+            Parameters:
+              Magnitude: *par7
+              Phase: *par8
             Amplitude:
               Class: SequentialAmplitude
               Amplitudes:
@@ -210,7 +208,9 @@ Intensity:
                     RecoilFinalState: 2
           - Class: CoefficientAmplitude
             Component: J/psi_-1_to_f2(1270)_-2+gamma_-1;f2(1270)_-2_to_pi0_0+pi0_0;
-            Parameters: [*par9, *par10]
+            Parameters:
+              Magnitude: *par9
+              Phase: *par10
             Amplitude:
               Class: SequentialAmplitude
               Amplitudes:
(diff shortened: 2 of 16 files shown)


redeboer:
> # [Codecov](https://codecov.io/gh/ComPWA/tensorwaves/pull/60?src=pr&el=h1) Report
> > Merging [#60](https://codecov.io/gh/ComPWA/tensorwaves/pull/60?src=pr&el=desc) into [master](https://codecov.io/gh/ComPWA/tensorwaves/commit/6c92ddf13fa083b10c39f759fc9e6571de3565d5&el=desc) will **decrease** coverage by `1.46%`.
> > The diff coverage is `23.80%`.
> 
> [Impacted file tree graph](https://codecov.io/gh/ComPWA/tensorwaves/pull/60?src=pr&el=tree)
> 
> ```diff
> @@            Coverage Diff             @@
> ##           master      #60      +/-   ##
> ==========================================
> - Coverage   55.66%   54.19%   -1.47%     
> ==========================================
>   Files           7        7              
>   Lines         318      334      +16     
> ==========================================
> + Hits          177      181       +4     
> - Misses        141      153      +12     
> ```
> 
> [Impacted Files](https://codecov.io/gh/ComPWA/tensorwaves/pull/60?src=pr&el=tree)	Coverage Δ	
> [tensorwaves/physics/helicityformalism/amplitude.py](https://codecov.io/gh/ComPWA/tensorwaves/pull/60/diff?src=pr&el=tree#diff-dGVuc29yd2F2ZXMvcGh5c2ljcy9oZWxpY2l0eWZvcm1hbGlzbS9hbXBsaXR1ZGUucHk=)	`27.01% <0.00%> (ø)`	
> [tensorwaves/physics/__init__.py](https://codecov.io/gh/ComPWA/tensorwaves/pull/60/diff?src=pr&el=tree#diff-dGVuc29yd2F2ZXMvcGh5c2ljcy9fX2luaXRfXy5weQ==)	`29.41% <25.00%> (-70.59%)`	
> [tensorwaves/interfaces.py](https://codecov.io/gh/ComPWA/tensorwaves/pull/60/diff?src=pr&el=tree#diff-dGVuc29yd2F2ZXMvaW50ZXJmYWNlcy5weQ==)	`100.00% <100.00%> (ø)`

We have to do something about the test coverage

tensorwaves/physics/__init__.py
```diff
@@ -1 +1,34 @@
...
+__all__ = ["helicityformalism"]
+
+import logging
+
+import yaml
+
 from . import helicityformalism
+
+
+def load_particle_list(particle_list_file: str) -> dict:
```
redeboer, line 12: All particle functionality may have to be migrated to the [expert system](https://github.com/ComPWA/expertsystem)

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -275,9 +275,9 @@ def __init__(self, recipe: list, builder):
             )
 
     def __call__(self, dataset):
-        seq_amp = tf.complex(atfi.const(1.0), atfi.const(0.0))
+        seq_amp = atfi.complex(atfi.const(1.0), atfi.const(0.0))
```
redeboer, line 278: Why use `amplitf` here?

tensorwaves/data/tf_phasespace.py
```diff
@@ -0,0 +1,58 @@
...
+    def __init__(
+        self, initial_state_mass: list, final_state_masses: list
+    ) -> None:
+        print(phasespace.__file__)
+        self.phsp_gen = phasespace.nbody_decay(
+            initial_state_mass, final_state_masses
+        )
+
+    def generate(
+        self, size: int, random_generator: UniformRealNumberGenerator
```
redeboer, line 27: This is what https://github.com/ComPWA/tensorwaves/issues/59 is about, right?

tensorwaves/data/tf_phasespace.py
```diff
@@ -0,0 +1,58 @@
...
+    ) -> np.darray:
+        weights, particles = self.phsp_gen.generate(n_events=size)
```
redeboer, line 29: `particles` or `four_momenta`?

tensorwaves/data/generate.py
```diff
@@ -0,0 +1,109 @@
+"""Tools to facilitate data sample generation."""
```
redeboer, line 1: Maybe more of a collection of façades?

tensorwaves/data/generate.py
```diff
@@ -0,0 +1,109 @@
...
+
+from tensorwaves.interfaces import (
+    Function,
+    Kinematics,
+    PhaseSpaceGenerator,
+    UniformRealNumberGenerator,
+)
+
+
+def _generate_data_bunch(
```
redeboer, line 17: What precisely is a bunch? Even though this method is private, a docstring may solve this ambiguity (although perhaps I miss some jargon here).

tensorwaves/data/generate.py
```diff
@@ -0,0 +1,109 @@
...
+    uniform_randoms = random_generator(bunch_size, max_value=maxvalue)
+
+    phsp_sample = phsp_sample.transpose(1, 0, 2)
+
+    return (phsp_sample[weights * intensities > uniform_randoms], maxvalue)
+
+
+def generate_data(
+    size: int,
+    phsp_generator: PhaseSpaceGenerator,
```
redeboer, line 40: May have to think of some structure to provide default arguments, same for `random_generator` for simpler scripts (as discussed). In that case, would also have to change the order of the arguments:
```python
def generate_data(
    size: int,
    intensity: Function,
    kinematics: Kinematics,
    phsp_generator: <?str or Enum?>,
    random_generator: <?UniformRealNumberGenerator, not sure how to provide default argument here>,
) -> np.ndarray:
```

tensorwaves/data/generate.py
```diff
@@ -0,0 +1,109 @@
...
+        if np.size(events) > 0:
+            events = np.vstack((events, bunch))
+        else:
+            events = bunch
+        progress_bar.next(np.size(bunch, 0))
+    progress_bar.finish()
+    return events[0:size].transpose(1, 0, 2)
+
+
+def generate_phsp(
```
redeboer, line 81: See comment at https://github.com/ComPWA/tensorwaves/pull/60/files#diff-1aaded47fc6bb3193bf9114d1f1ed943R40

tensorwaves/data/generate.py
```diff
@@ -0,0 +1,109 @@
...
+    phsp_generator: PhaseSpaceGenerator,
+    random_generator: UniformRealNumberGenerator,
+    intensity: Function,
+    kinematics: Kinematics,
+) -> np.ndarray:
+    """Create a data sample based on an intensity."""
+    events = np.array([])
+
+    current_max = 0.0
+    bunch_size = 50000
```
redeboer, line 49: Maybe give user the option to specify `bunch_size` through function argument?

tensorwaves/data/generate.py
```diff
@@ -0,0 +1,109 @@
...
+
+def generate_phsp(
+    size: int,
+    phsp_generator: PhaseSpaceGenerator,
+    random_generator: UniformRealNumberGenerator,
+) -> np.ndarray:
+    """Create a phase space sample."""
+    events = np.array([])
+
+    bunch_size = 50000
```
redeboer, line 89: https://github.com/ComPWA/tensorwaves/pull/60/files#diff-1aaded47fc6bb3193bf9114d1f1ed943R49

tensorwaves/data/generate.py
```diff
@@ -0,0 +1,109 @@
...
+
+    progress_bar = Bar("Generating", max=size, suffix="%(percent)d%%")
+
+    while np.size(events) < size:
+        particles, weights = phsp_generator.generate(
+            bunch_size, random_generator
+        )
+        particles = particles.transpose(1, 0, 2)
+
+        hit_and_miss_randoms = random_generator(bunch_size)
```
redeboer, line 99: I suppose this is also not deterministic, but don't know how to do this otherwise.

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: Oh, I was just testing something. Ideally we want all of the calls to be amplitf calls, because there it is foreseen to have a tensorflow or numpy backend, which can be chosen dynamically. So by using the amplitf code you are flexible in chosing the backend. But all of the tf calls would have to be swapped by amplitf calls and amplitf also has to be extended.

tensorwaves/data/tf_phasespace.py
spflueger: yes

tensorwaves/data/generate.py
spflueger: Hmm maybe chunk is a better word. Its only a small fix size sample. So depending on the requested total sample size, many of those bunches have to be created and concatenated. I can add a docstring though

tensorwaves/data/generate.py
spflueger: true

tensorwaves/data/generate.py
spflueger: it should be deterministic, or why you assume its not?

examples/intensity-recipe.yaml
```diff
@@ -82,7 +72,9 @@ Intensity:
         Amplitudes:
           - Class: CoefficientAmplitude
             Component: J/psi_-1_to_f2(1270)_0+gamma_-1;f2(1270)_0_to_pi0_0+pi0_0;
-            Parameters: [*par1, *par2]
+            Parameters:
+              Magnitude: *par1
```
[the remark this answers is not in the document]
spflueger: Can you clarify the type at part? I don't understand which part you are referring to

tensorwaves/physics/__init__.py
spflueger: Yes, I think once the expert system is compatible with tensorwaves we can migrate this code

examples/intensity-recipe.yaml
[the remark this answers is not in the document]
redeboer: Seems the link is broken, it refered to line 20 of the `intensity-recipe.yaml` file
https://github.com/ComPWA/tensorwaves/blob/1d3436287598a9f4cec7f60478d905e26526beae/examples/intensity-recipe.yaml#L20

tensorwaves/data/generate.py
redeboer: Because `[particles[weights > hit_and_miss_randoms]` later on seems like a random process to me

examples/test.py
```diff
@@ -1,10 +1,44 @@
...
+    recipe = yaml.load(fc.read(), Loader=yaml.SafeLoader)
+    kin = HelicityKinematics(
+        {
+            x["ID"]: pos
+            for pos, x in enumerate(recipe["Kinematics"]["FinalState"])
+        }
+    )
+    part_list = load_particle_list("examples/intensity-recipe.yaml")
+
+    generator = TFPhaseSpaceGenerator(JPSI_MASS, [0.0, PI0_MASS, PI0_MASS])
```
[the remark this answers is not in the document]
spflueger: The current code does this for the `HelicityKinematics`. The PhaseSpaceGenerator then uses the Kinematics instance for initialization

examples/intensity-recipe.yaml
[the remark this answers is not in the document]
spflueger: Good point, but this is completely unrelated to this. I opened a new issue for that #61

tensorwaves/data/generate.py
spflueger: The current code addresses this issue. However we might have to think about this a bit more, as the dependency injection we use currently assumes that all constructors of phase space generators and random number generators are the same (which does not have to be true)

spflueger:
Ok, I made some more improvements (see https://github.com/ComPWA/tensorwaves/pull/60/commits/6f304480b215b9d0e5d1d8363193fe7105232685)
For me this would wrap up this PR.

codecov.yml
```diff
@@ -4,13 +4,13 @@ codecov:
 coverage:
   precision: 2
   round: down
-  range: "75...100"
+  range: "60...90" # color range
```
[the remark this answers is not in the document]
spflueger: really? https://docs.codecov.io/docs/codecovyml-reference#coveragerange

examples/test.py
```diff
@@ -1,10 +1,32 @@
...
+
+    builder = IntensityBuilder(part_list, kin, phsp_sample)
+    intensity = builder.create_intensity(recipe)
+
+    data_sample = generate_data(30000, kin, intensity)
+
+    dataset = kin.convert(data_sample)
+
+    # ploting
+    import pandas as pd
```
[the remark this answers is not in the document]
spflueger: Yeah, this was my goal after the next PR.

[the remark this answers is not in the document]
redeboer: https://github.com/ComPWA/tensorwaves/issues/62

codecov.yml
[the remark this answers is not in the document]
redeboer: Sorry I should specify: the `target` has been changed to 60 as well, which means we have lowered our coverage standards.

tensorwaves/physics/particle.py
```diff
@@ -0,0 +1,33 @@
+"""Collection of helper functions for particles."""
```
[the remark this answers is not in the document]
spflueger: Hmm, I was also thinking about this part. I think this deserves its own issue. In my current opinion it already depends on how we want this particle functionality to look like. If we use a dict as the data structure and just load and save it to various formats, all you need is some "specification" on what keys and structure you have inside the dict (current situation).
If we choose to have some separate module, only these two options make sense:
1. Put all particle functionality into expert-system. As it needs this and is lower level. So the recipes generated with the expert system have to be readable by tensorwaves.
2. Have a separate particle repository which can be used

tensorwaves/data/__init__.py
```diff
@@ -0,0 +1,4 @@
+__all__ = ["generate", "tf_phasespace"]
```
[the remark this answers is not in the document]
spflueger: Makes sense, I would fix that in the next PR that is lined up.

tensorwaves/physics/particle.py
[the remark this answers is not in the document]
redeboer: Perhaps this issue is even more general: both `expertsystem` and `tensorwaves` (perhaps even `pycompwa`) need to agree on the syntax on the recipe, so it's natural to put some sort of recipe module into the expert system.

tests/data/test_generate.py
```diff
@@ -0,0 +1,54 @@
+"""Tests the generate facade functions."""
```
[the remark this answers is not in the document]
spflueger: I also thought about that. It was still a little useful though. So we have to think about which tools to exclude for tests

PR #63, opened 2020-05-08
feat!: implement Minuit2 optimization functionality 

spflueger:


tensorwaves/estimator.py
```diff
@@ -1,11 +1,49 @@
...
+        self.__model = model
+        self.__dataset = dataset
+
+    def __call__(self) -> float:
+        props = self.__model(self.__dataset)
+        logs = tf.math.log(props)
+        log_lh = tf.reduce_sum(logs)
+        return -log_lh.numpy()
+
+    def gradient(self) -> np.ndarray:
```
redeboer, line 31: There's no issue on the missing gradient yet, right?

tensorwaves/estimator.py
```diff
@@ -1,11 +1,49 @@
...
-        super(UnbinnedLL, self).__init__(name="UnbinnedLL")
+    @property
+    def parameters(self) -> dict:
+        return self.__model.parameters
 
-    def call(self, y_true, y_pred):
-        logs = tf.math.log(y_pred)
-        ll = tf.reduce_sum(logs)
-        return ll
+    def update_parameters(self, new_parameters: dict) -> None:
```
redeboer, line 48: Why not a setter?

tensorwaves/optimizer/minuit.py
```diff
@@ -0,0 +1,71 @@
+"""Minuit2 adapter to the iminuit package."""
```
redeboer, line 1: Intersphinx reference to `iminuit`'s API?
https://iminuit.readthedocs.io/en/latest/

tensorwaves/estimator.py
spflueger: To me the `update_parameters` makes it clearer that the you are not replacing parameters but merely updating some of their values

tensorwaves/optimizer/minuit.py
spflueger: True

tensorwaves/estimator.py
spflueger: Good point. I will open an issue and remove this code.

spflueger: See #64

PR #71, opened 2020-05-28
feat: add Blatt-Weisskopf barrier factor option

### LINKED ISSUE #51: Add barrier factors functionality
- [x] retrieve appropriate kinematic variables (invariant masses) **[Estimated effort: 1h]**
- [ ] define new call wrapper and use amplitf breit_wigner_lineshape function **[Estimated effort: 1h]**

spflueger:
Closes #51

### DIFF
--- a/examples/intensity-recipe.yaml
+++ b/examples/intensity-recipe.yaml
@@ -1258,20 +1258,31 @@ ParticleList:
 
 Dynamics:
   J/psi:
-    Type: nonDynamic
+    Type: NonDynamic
+    FormFactor:
+      Type: BlattWeisskopf
+      MesonRadius: { Value: 1.0 }
 
   f0(980):
-    Type: relativisticBreitWigner
-    FormFactor: BlattWeisskopf
+    Type: RelativisticBreitWigner
+    FormFactor:
+      Type: BlattWeisskopf
+      MesonRadius: { Value: 1.0 }
 
   f2(1270):
-    Type: relativisticBreitWigner
-    FormFactor: BlattWeisskopf
+    Type: RelativisticBreitWigner
+    FormFactor:
+      Type: BlattWeisskopf
+      MesonRadius: { Value: 1.0 }
 
   f0(1500):
-    Type: relativisticBreitWigner
-    FormFactor: BlattWeisskopf
+    Type: RelativisticBreitWigner
+    FormFactor:
+      Type: BlattWeisskopf
+      MesonRadius: { Value: 1.0 }
 
   f2(1950):
-    Type: relativisticBreitWigner
-    FormFactor: BlattWeisskopf
+    Type: RelativisticBreitWigner
+    FormFactor:
+      Type: BlattWeisskopf
+      MesonRadius: { Value: 1.0 }
--- a/tensorwaves/physics/helicityformalism/amplitude.py
+++ b/tensorwaves/physics/helicityformalism/amplitude.py
@@ -8,14 +8,24 @@ callable.
 
 import logging
 from collections import namedtuple
-from typing import Any, Callable, Dict, List, Tuple
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+)
 
 import amplitf.interface as atfi
 from amplitf.dynamics import (
     blatt_weisskopf_ff_squared,
     relativistic_breit_wigner,
 )
-from amplitf.kinematics import two_body_momentum, wigner_capital_d
+from amplitf.kinematics import (
+    two_body_momentum,
+    wigner_capital_d,
+)
 
 import numpy
 
@@ -150,7 +160,9 @@ class IntensityBuilder:
         decay_dynamics = self._dynamics[decaying_state_name]
         kwargs = {}
         if "FormFactor" in decay_dynamics:
-            kwargs.update({"FormFactor": decay_dynamics["FormFactor"]["Type"]})
+            kwargs.update(
+                {"form_factor": decay_dynamics["FormFactor"]["Type"]}
+            )
             meson_radius_par = self.register_parameter(
                 "MesonRadius_" + decaying_state_name,
                 decay_dynamics["FormFactor"]["MesonRadius"]["Value"],
@@ -393,13 +405,14 @@ DynamicsProperties = namedtuple(
 
 class _RelativisticBreitWigner:
     def __init__(
-        self, dynamics_props: DynamicsProperties, **kwargs: dict
+        self,
+        dynamics_props: DynamicsProperties,
+        form_factor: Optional[str] = None,
     ) -> None:
         self._dynamics_props = dynamics_props
         self._call_wrapper = self._without_form_factor
-        if "FormFactor" in kwargs:
-            if kwargs["FormFactor"] == "BlattWeisskopf":
-                self._call_wrapper = self._with_form_factor
+        if form_factor == "BlattWeisskopf":
+            self._call_wrapper = self._with_form_factor
 
     def __call__(self, dataset: dict) -> tf.Tensor:
         return self._call_wrapper(dataset)
@@ -432,15 +445,16 @@ class _RelativisticBreitWigner:
 
 class _NonDynamic:
     def __init__(
-        self, dynamics_props: DynamicsProperties, **kwargs: dict
+        self,
+        dynamics_props: DynamicsProperties,
+        form_factor: Optional[str] = None,
     ) -> None:
         self._dynamics_props = dynamics_props
         self._call_wrapper: Callable[
             [dict], tf.Tensor
         ] = self._without_form_factor
-        if "FormFactor" in kwargs:
-            if kwargs["FormFactor"] == "BlattWeisskopf":
-                self._call_wrapper = self._with_form_factor
+        if form_factor == "BlattWeisskopf":
+            self._call_wrapper = self._with_form_factor
 
     def __call__(self, dataset: dict) -> tf.Tensor:
         return self._call_wrapper(dataset)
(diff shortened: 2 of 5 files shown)


tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -118,7 +118,12 @@ class SubSystem:
     * :attr:`parent_recoil_state`
     """
 
-    def __init__(self, final_states, recoil_state, parent_recoil_state):
+    def __init__(
+        self,
+        final_states: Union[Tuple, List],
```
redeboer, line 123: I think `typing` recommends [`Sequence`](https://docs.python.org/3/library/typing.html#typing.Sequence) here

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -158,7 +165,7 @@ def __eq__(self, other):
             return False
         return True
 
-    def __hash__(self):
+    def __hash__(self) -> int:
```
redeboer, line 168: May want to derive this class from [`abc.Hashable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Hashable)

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -199,19 +206,19 @@ def reaction_kinematics_info(self) -> ParticleReactionKinematicsInfo:
...
 
         Return:
             `float`
         """
         return 1.0
 
-    def is_within_phase_space(self, events):
+    def is_within_phase_space(self, events: np.ndarray) -> Tuple[bool]:
         """Check whether events lie within the phase space definition."""
         raise NotImplementedError("Currently not implemented.")
```
redeboer, line 219: Remove the string, the exception class is already clear by itself.

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -199,19 +206,19 @@ def reaction_kinematics_info(self) -> ParticleReactionKinematicsInfo:
         return self._reaction_info
 
     @property
-    def phase_space_volume(self):
+    def phase_space_volume(self) -> float:
         """Get volume of the defined phase space.
 
         Return:
             `float`
```
redeboer, line 213: Can be removed because of the type hint

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -12,7 +12,7 @@
 :class:`~SubSystem`.
 """
 import logging
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
```
redeboer, line 15: I start to prefer:
```python
from typing import (
    Dict,
    List,
    Optional,
    Tuple,
    Union,
)
```
because of diffs and easier sorting (as required by pylint). Note that `black` falls to this format once you wrap this list in parentheses, ending it with a commit.

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -294,7 +303,7 @@ def _convert_ids_to_indices(self, ids: tuple):
 
         return ids
 
-    def convert(self, events):
+    def convert(self, events: np.ndarray) -> dict:
```
redeboer, line 306: > This is also known as a dataset

...suggests that we need to create some object for such a dataset, not sticking with a `dict`. Maybe make an issue for this? Might be addressed by [#23](https://github.com/ComPWA/tensorwaves/issues/23) though.

examples/intensity-recipe.yaml
```diff
@@ -1258,20 +1258,31 @@ ParticleList:
 
 Dynamics:
   J/psi:
-    Type: nonDynamic
+    Type: NonDynamic
+    FormFactor:
+      Type: BlattWeisskopf
+      MesonRadius: { Value: 1.0 }
```
redeboer, line 1264: Can a meson radius have an uncertainty?

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -136,31 +139,27 @@ def create_element(self, recipe: dict) -> Callable:
...
         dynamics_properties: "DynamicsProperties",
     ) -> Callable:
         """Create a dynamics function callable."""
         dynamics_builder = self._get_dynamics_builder(decaying_state_name)
-        return dynamics_builder(dynamics_properties)
+
+        decay_dynamics = self._dynamics[decaying_state_name]
+        kwargs = {}
+        if "FormFactor" in decay_dynamics:
+            kwargs.update({"FormFactor": decay_dynamics["FormFactor"]["Type"]})
```
redeboer, line 153: Does the key `"FormFactor"` agree with PEP conventions? If I understand `kwargs` correctly, this means one of the arguments of the callable will be `FormFactor`, not `formfactor`.

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -387,46 +386,84 @@ def _create_sequential_amplitude(
         "inv_mass_name",
         "inv_mass_name_prod1",
         "inv_mass_name_prod2",
+        "meson_radius",
```
redeboer, line 389: I think we should switch to [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) as it allows for much more powerful syntax than `collections.namedtuple`. Create an issue for this?

Note that we already enfoce Python >3.6, so that should be fine.

examples/intensity-recipe.yaml
spflueger: yeah, it can be a fit parameter just like mass

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: Yeah, can you create an issue for that and we deal with that separately

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: good point

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Hmm, yeah I think we can discuss this in #23 as its related to this.

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: I improved the code further. Check it out and resolve if you are satisfied

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: done

examples/intensity-recipe.yaml
spflueger: resolve it, if you are ok with it

tensorwaves/physics/helicityformalism/kinematics.py
redeboer: Sorry, actually I was referring to `NotImplementedError("Currently not implemented.")` -- the class `NotImplementedError` does not have to be constructed with a string.

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -12,7 +12,14 @@
 :class:`~SubSystem`.
 """
 import logging
-from typing import List, Optional, Union
+from collections.abc import Hashable
```
redeboer, line 15: Small thought: since [`typing.Hashable`](https://docs.python.org/3/library/typing.html#typing.Hashable) is not the same as [`abc.Hashable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Hashable) (this is the case for many typing vs abstract base metaclasses), I tend to use `from collections import abc` and use `abc.WhateverClass`, then `from typing import WhateverClass` for the actual typing, so that there is no conflict between the two and the intension is clear from the namespace. It doesn't really matter here, because we only use `abc.Hashable` to enforce `__hash__` definition, so consider this a general thought.

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -274,15 +284,15 @@ def register_subsystem(self, subsystem: SubSystem):
...
         invmass_name = self.register_invariant_mass(list(set(state_fs)))
         angle_names = self.register_helicity_angles(subsystem)
 
         return (invmass_name,) + angle_names
 
-    def _convert_ids_to_indices(self, ids: tuple):
+    def _convert_ids_to_indices(self, ids: tuple) -> Sequence:
         """Convert unique ids to event indices.
 
         Uses the :attr:`_fs_id_event_pos_mapping`.
```
redeboer, line 298: Interesting, does this render in the API (and does `linkcheck` complain if there's a typo) for 'private' attributes as well?

tensorwaves/physics/helicityformalism/kinematics.py
```diff
@@ -274,15 +284,15 @@ def register_subsystem(self, subsystem: SubSystem):
...
+        state_fs: list = []
         for fs_uid in subsystem.final_states:
             state_fs += fs_uid
         invmass_name = self.register_invariant_mass(list(set(state_fs)))
         angle_names = self.register_helicity_angles(subsystem)
 
         return (invmass_name,) + angle_names
 
-    def _convert_ids_to_indices(self, ids: tuple):
+    def _convert_ids_to_indices(self, ids: tuple) -> Sequence:
```
redeboer, line 295: Here, I guess the output should be `Tuple` again, as it's the same as the input.

Generally, I have the impression that `Sequence` is used if you for instance don't know the whether the input is `List` or `Tuple`, but you don't use `Sequence` for output if you have control over the output.

spflueger: Yeah, I'm aware of that. That function is return either one or the other though. But let me see if I can change that.

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Good point. Since its not used in the docs, it also did not throw any errors on mistakes in the docstring. So I removed the docstring completely

tensorwaves/physics/helicityformalism/kinematics.py
spflueger: Ok changed to tuple now

Issue #73, opened 2020-06-17
Illustrate how to use the expertsystem with tensorwaves for the helicity formalism


redeboer:
With https://github.com/ComPWA/expertsystem/commit/1d822590d42cfdae2af380f766867aa1b3140c46, the expertsystem can write amplitude models for the helicity formalism to a YAML format that tensorwaves understand. So now we run tensorwaves over arbitrary decays with the helicity formalism (so far we just use one [`intensity-recipe.yaml`](https://github.com/ComPWA/tensorwaves/blob/e515837ac740af940ac846947e85924a2179afdb/examples/intensity-recipe.yaml) file as an example/benchmark).

Note: there are tiny improvements to the YAML recipe file format, so that will have to be fixed in tensorwaves.

redeboer:
Best to first merge https://github.com/ComPWA/expertsystem/pull/51 so that we can create a release that has a better README, see https://github.com/ComPWA/expertsystem/pull/51#issue-430928105

PR #74, opened 2020-06-19
docs: add workflow step create amplitude model

redeboer:
Based on
https://pwa.readthedocs.io/projects/expertsystem/en/0.1.2-alpha/usage/QuickStart.html


redeboer:
@spflueger If you want to keep this PR small, we can keep it with point 1 (the create model notebook only) and leave point 2 (use output recipe in generate notebook) to a follow-up PR. Thing is that this will require small modifications to the codebase itself

redeboer:
@spflueger Seeing that it will take a bit more effort to adapt tensorwaves so that it takes the (slightly modified) YAML recipe file that the expertsystem produces (version 0.1.2a0), I think it's better to leave that up to a follow-up PR that can close #73.

So this one is only to have a notebook that creates the recipe, the generate notebook still uses the existing `intensity-recipe.yaml`.

PR #76, opened 2020-06-19
refactor: switch to typing.NamedTuple

### LINKED ISSUE #72: Switch from collections.namedtuple to typing.NamedTuple
I think we should switch to [`typing.NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) as it allows for much more powerful syntax than `collections.namedtuple`. Create an issue for this?

Note that we already enfoce Python >3.6, so that should be fine.

_Originally posted by @redeboer in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjcwMTg5OTE3OnYy/pull_request_review_threads/discussion_

redeboer:
Closes #72

### DIFF
--- a/tensorwaves/physics/helicityformalism/amplitude.py
+++ b/tensorwaves/physics/helicityformalism/amplitude.py
@@ -7,7 +7,6 @@ callable.
 """
 
 import logging
-from collections import namedtuple
 from typing import (
     Any,
     Callable,
@@ -16,6 +15,7 @@ from typing import (
     Optional,
     Tuple,
 )
+from typing import NamedTuple
 
 import amplitf.interface as atfi
 from amplitf.dynamics import (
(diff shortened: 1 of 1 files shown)


tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -385,22 +385,24 @@ def _create_sequential_amplitude(
...
+class DynamicsProperties(NamedTuple):
+    """Data structure representing dynamic properties."""
+
+    orbit_angular_momentum: float
+    resonance_mass: float
+    resonance_width: float
+    inv_mass_name: str
+    inv_mass_name_prod1: str
+    inv_mass_name_prod2: str
+    meson_radius: Optional[float]
```
redeboer, line 405: @spflueger The key of this PR is whether you agree with the types here (that's one of the advantages of using `typing.NamedTuple` over `collections.namedtuple`). I wasn't 100% here as I didn't write this module.

spflueger: looks good to me

Issue #78, opened 2020-06-22
Enforce pre-commit checks through Travis


redeboer:
While working on the import of the new recipe file structure, I found that there are still several linting errors in the code. This can be prevented by enforcing all pre-commit checks upon pull requests. In addition, it would help to make the checks performed by pre-commit consistently appear in VScode.

Of course, those remaining linting errors will have to be fixed and some of the config files for e.g. pylint have to be revisited.

A step plan:
- [x] Add VScode folder configuration (see https://github.com/ComPWA/expertsystem/pull/41)
- [x] Add [`pydocstyle`](https://pypi.org/project/pydocstyle/). We did this in https://github.com/ComPWA/expertsystem/pull/28 and it worked quite well along with `flake8` and `pylint` since.
- [x] Fix `flake8` errors and remove conflicts with other linters
- [x] Fix `mypy` errors
- [x] Fix `pylint` errors and fix config problems for external packages
- [x] Add [`rstcheck`](https://github.com/myint/rstcheck#configuration-file) (used in VScode, but available as pre-commit hook as well))
- [x] Add [`prettier`](https://prettier.io/) formatter for Markdown, as this appears to be the only tool that can hard-wrap lines through pre-commit
- [x] Enforce pre-commit through Travis CI on _all_ files (`pre-commit run -a`)
- [x] Explain the importance of linting tools in the contributing section (i.e., they formulate and automatise our convention policies)

What's nice about this setup is that new checks can be easily implemented as a new pre-commit hook and then they are immediately enforced through Travis CI as well without having to adapt `.travis.yml` as well.

PR #79, opened 2020-06-22
ci: standardize VSCode settings

redeboer:
Part 1 of #78 

For the prior discussion, see https://github.com/ComPWA/expertsystem/pull/41


redeboer:
> And there the diffs have line wrapping

Oh that's interesting! But I also saw there are some bugs there, like, the emojis in labels don't render correctly

PR #81, opened 2020-06-22
ci!: add and enforce pydocstyle check

redeboer:
Part 3 of #78 

See also https://github.com/ComPWA/expertsystem/pull/28

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,12 @@ repos:
         language: system
         types:
           - python
+      - id: pydocstyle
+        name: pydocstyle
+        entry: pydocstyle
+        language: system
+        types:
+          - python
       - id: pylint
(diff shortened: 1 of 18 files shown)


.pre-commit-config.yaml
```diff
@@ -34,6 +34,12 @@ repos:
         language: system
         types:
           - python
+      - id: pydocstyle
+        name: pydocstyle
+        entry: pydocstyle
+        language: system
+        types:
+          - python
```
spflueger, line 42: Why we need this again? Because flake8 cant catch all errors?

redeboer: Initially, yes https://github.com/ComPWA/expertsystem/pull/28#pullrequestreview-411084795

But I found pydocstyle is also way better with Google (`napoleon`) conventions

PR #82, opened 2020-06-22
ci!: enforce flake8 through Travis

redeboer:
Part 4 of #78

### DIFF
--- a/examples/python/workflow.py
+++ b/examples/python/workflow.py
@@ -2,19 +2,26 @@
 
 import logging
 
+import matplotlib.pyplot as plt
+
+import pandas as pd
+
 import yaml
 
 from tensorwaves.data.generate import generate_data, generate_phsp
 from tensorwaves.estimator import UnbinnedNLL
+from tensorwaves.optimizer.minuit import Minuit2
 from tensorwaves.physics.helicityformalism.amplitude import IntensityBuilder
 from tensorwaves.physics.helicityformalism.kinematics import HelicityKinematics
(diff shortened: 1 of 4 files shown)


examples/python/workflow.py
```diff
@@ -66,12 +69,16 @@
...
-    for name, value in estimator.parameters.items():
+    for name in estimator.parameters:
         if name in free_params:
             params[name] = free_params[name]
     logging.info(params)
 
     logging.info("starting fit")
-    minu2 = Minuit2()
-    result = minu2.optimize(estimator, params)
+    minuit2 = Minuit2()
```
redeboer, line 78: Sorry this isn't a flake error, but popped up in cSpell, apparently

tox.ini
```diff
@@ -25,13 +26,23 @@ exclude =
 ignore = # more info: https://www.flake8rules.com/
     D102 # method docstring
     D103 # function docstring
+    D105 # docstring in magic method
     D107 # init docstring
+    D400 # method docstring
```
spflueger, line 31: This is first line ending with a period or? Why that has to be ignored?

redeboer: Actually I think we can get rid of the D warnings under flake by just removing flake8-docstring. Let me give that a try

redeboer: `flake8-docstrings` indeed causes these warnings and `pydocstyle` makes them redundant 45e473d9932998ec0b965a3257713db2c6f769b7

Issue #54, opened 2020-04-30
Add mypy check to Travis CI


redeboer:
With the implementation of a `mypy` pre-commit (https://github.com/ComPWA/tensorwaves/pull/48), it may be worth it to add the check to Travis CI as well. That would just be adding
```bash
mypy --config-file tox.ini .
```
under this line https://github.com/ComPWA/tensorwaves/blob/effe261f9048cd27ddead086eb8a499450b93ca7/.travis.yml#L41

It would require to add hints to all existing code (excluding the `expertsystem`).

PR #83, opened 2020-06-22
ci!: enforce mypy check through Travis

redeboer:
Part 5 of #78 

Closes https://github.com/ComPWA/tensorwaves/issues/54

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
           - python
       - id: pylint
         name: pylint
-        entry: pylint --rcfile=.pylintrc
+        entry: pylint
         language: system
         types:
           - python
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
           - python
       - id: pylint
         name: pylint
-        entry: pylint
+        entry: pylint --rcfile=.pylintrc
         language: system
         types:
           - python
--- a/tensorwaves/data/generate.py
+++ b/tensorwaves/data/generate.py
@@ -18,6 +18,7 @@ from tensorwaves.interfaces import (
     UniformRealNumberGenerator,
 )
 from tensorwaves.physics.helicityformalism.kinematics import (
+    HelicityKinematics,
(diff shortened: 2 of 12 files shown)


redeboer:
@spflueger So here things become a bit more tricky with regard to `Kinematics` --> `HelicityKinematics`. But I think this illustrates we have a problem with regard to the design here.

.pre-commit-config.yaml
```diff
@@ -42,7 +42,7 @@ repos:
           - python
       - id: pylint
         name: pylint
-        entry: pylint
+        entry: pylint --rcfile=.pylintrc
```
redeboer, line 45: Hm this one should be part of the next PR

tensorwaves/data/generate.py
```diff
@@ -45,7 +46,7 @@ def _generate_data_bunch(
 
 def generate_data(
     size: int,
-    kinematics: Kinematics,
+    kinematics: HelicityKinematics,
```
spflueger, line 49: Hmm, small comment. Originally the idea was to have this generate function work generally, independent on the Kinematics. However the `PhaseSpaceGenerator`s are working based on particle transition info.
I was thinking about duck typing here. That the specific Kinematics instance would have to define this attribute for the generate function to work w/o errors... Is a bit of a code smell though. But I also wanted the interface to be simple. Maybe you have a good idea for this problem.

So I guess for now we can go with `HelicityKinematics`, and create some Issue for this general Kinematics option.

spflueger:
> @spflueger So here things become a bit more tricky with regard to `Kinematics` --> `HelicityKinematics`. But I think this illustrates we have a problem with regard to the design here.

Yeah I agree. Also made a small comment in my review.

tensorwaves/data/generate.py
redeboer: Yes, that's true and I understood this was done with duck typing in mind. Did make me wonder what typing and mypy thinks about duck typing. I'm inclined to avoid duck typing in combination with type hinting.

As for our design: I initially gave a try adding the transition info into the `Kinematics` base class, but that changed the design too radically.

Issue #77, opened 2020-06-19
Modify amplitude and kinematics modules so that they can handle the new recipe file structure


redeboer:
Follow-up to #74 
Required for #73 

The expertsystem v0.1.2a0 produces a slightly different amplitude model recipe file that the hard-coded `intensity-recipe.yaml` file.

spflueger:
I believe this refers to the problem, that a predefined recipe structure is assumed in the kinematics code as well right?

PR #84, opened 2020-06-22
ci!: enforce pylint check through Travis

redeboer:
Part 6 of #78 

Required some refactoring because of `too-many-locals`

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -23,4 +23,5 @@ ignored-modules=
 # Minimum number of public methods for a class (see R0903). Default is 2.
 # We decrease this to 1, since some interface classes just have a single
 # (public) method
+max-args=7  # default: 5
 min-public-methods=1
--- a/.pylintrc
+++ b/.pylintrc
@@ -20,8 +20,8 @@ ignored-modules=
     amplitf.interface,
 
 [DESIGN]
+max-args=7  # default: 5
 # Minimum number of public methods for a class (see R0903). Default is 2.
 # We decrease this to 1, since some interface classes just have a single
 # (public) method
-max-args=7  # default: 5
 min-public-methods=1
--- a/.pylintrc
+++ b/.pylintrc
@@ -6,7 +6,11 @@
 init-import=yes
 
 [MESSAGES CONTROL]
-disable=C0116, C0330, RST203, RST301
+disable=
+    C0116,  # missing-function-docstring
+    C0330,  # for black formatting
+    RST203,
+    RST301,
 
 [MASTER]
 ignore=conf.py
--- a/.pylintrc
+++ b/.pylintrc
@@ -15,6 +15,10 @@ disable=
 [MASTER]
 ignore=conf.py
 
+[TYPECHECK]
+ignored-modules=
+    amplitf.interface,
+
 [DESIGN]
 # Minimum number of public methods for a class (see R0903). Default is 2.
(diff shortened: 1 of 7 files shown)


.pylintrc
```diff
@@ -6,13 +6,22 @@
...
 
+[TYPECHECK]
+ignored-modules=
+    amplitf.interface,
+
 [DESIGN]
 # Minimum number of public methods for a class (see R0903). Default is 2.
 # We decrease this to 1, since some interface classes just have a single
 # (public) method
+max-args=7  # default: 5
```
spflueger, line 26: can you shift the comment down

redeboer: Which comment? `# default`?

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -501,56 +504,104 @@ def __call__(self, dataset: dict) -> tf.Tensor:
...
+            recoil_system.recoil_ids = _safe_wrap_list(
+                definition["RecoilSystem"]["RecoilFinalState"]
+            )
+        if "ParentRecoilFinalState" in definition["RecoilSystem"]:
+            recoil_system.parent_recoil_ids = _safe_wrap_list(
+                definition["RecoilSystem"]["ParentRecoilFinalState"]
+            )
+        return recoil_system
+
+
```
redeboer, line 561: So this part goes a bit into #77. It doesn't fix that issue, but it's an attempt at formulating a data structure that can then be filled/created from a recipe file (`dict`).

Anyway, in this PR the motivation for creating these classes was just to reduce the number of local variables under `_create_helicity_decay` that `pylint` was complaining about.

tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -501,56 +504,104 @@ def __call__(self, dataset: dict) -> tf.Tensor:
...
+    ) -> None:
+        self.recoil_ids: List[int] = []
+        self.parent_recoil_ids: List[int] = []
+        if recoil_ids:
+            self.recoil_ids = recoil_ids
+        if parent_recoil_ids:
+            self.parent_recoil_ids = parent_recoil_ids
+
+    @staticmethod
+    def from_dict(definition: Dict[str, Any]) -> "_RecoilSystem":
```
spflueger, line 549: I guess those changes where necessary to get rid of the complexity issue from pylint?

redeboer: https://github.com/ComPWA/tensorwaves/pull/84#discussion_r443465116

.pylintrc
spflueger: The comment above corresponds to the line below.

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: Oh sorry, so that comment too late. All right.

.pylintrc
redeboer: Oops yes

redeboer:
Hmm so this shows we need to write more tests...
https://travis-ci.com/github/ComPWA/tensorwaves/jobs/352148558

.pylintrc
redeboer: 7bb4b205900ab26f37c27e0ad11ee0eef0cd4ac5

PR #88, opened 2020-06-22
feat: add support for canonical formalism

### LINKED ISSUE #52: Add canonical formalism functionality
- [x] retrieve Clebsch Gordan factor from the recipe **[Estimated effort: 10min]**
- [x] evaluate the factor into a constant **[Estimated effort: 1h-1d]**
- [x] set L appropriately **[Estimated effort: 1min]**

spflueger:
* Add support for canonical sum from a helicity amplitude. This defines
the appropriate Clebsch Gordan coefficients.
* Add tests for canonical formalism parts.

Closes #52


tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -29,6 +29,8 @@
 
 import numpy
 
+from sympy.physics.quantum.cg import CG
```
redeboer, line 32: Should be added to `requirements.txt` and `setup.py`, right?

tests/physics/helicityformalism/test_canonical.py
```diff
@@ -0,0 +1,176 @@
...
+                        "J": 1.0,
+                        "M": 1.0,
+                        "j1": 0.0,
+                        "m1": 0.0,
+                        "j2": 1.0,
+                        "m2": 1.0,
+                    }
+                },
+            },
+            math.sqrt(1 / 10) * 1,
```
redeboer, line 101: Hmm class `dict` problem again... Keep thinking there's a case for https://github.com/ComPWA/expertsystem/issues/22

spflueger: Hmm, but the current code just works with dicts? So this is unrelated comment right?

redeboer: Yeah, it's not something for the review.

requirements.txt
```diff
@@ -4,5 +4,6 @@ numpy
 phasespace
 progress
 pyyaml
+sympy
```
redeboer, line 7: Should also be added to `setup.py` (which is used by PyPI)

requirements.txt
```diff
@@ -4,5 +4,6 @@ numpy
...
 tensorflow==2.1.*
 xmltodict
```
redeboer, line 9: `xmltodict` can indeed be removed (also from `setup.py`!)

requirements.txt
spflueger: 1bf2ee2ba9cb0527d925a6388fa2b82a835e51e2

requirements.txt
spflueger: 1bf2ee2ba9cb0527d925a6388fa2b82a835e51e2

spflueger:
> Nice! Now all we need is to illustrate how to use all this in some notebooks (and ideally in the tests as to keep coverage high).

Yes, I think one the expertsystem output can be parsed directly we can start working on the notebooks etc

PR #92, opened 2020-06-25
ci: make pydocstyle work in vscode

redeboer:
`pydocstyle` linting in vscode didn't work correctly, probably because of the `pydocstyle` args provided through `.vscode/settings.json`:
https://github.com/ComPWA/tensorwaves/blob/0c8c706f268d127c2ecfaeeb7b06cab22daf00b8/.vscode/settings.json#L26
This PR introduces another way to ignore docstrings in the tests folder that works for both linting from the terminal and in vscode.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -7,7 +7,9 @@ init-import=yes
 
 [MESSAGES CONTROL]
 disable=
-    C0116,  # missing-function-docstring
+    C0114,  # done by pydocstyle
+    C0115,  # done by pydocstyle
+    C0116,  # done by pydocstyle
     C0330,  # for black formatting
     RST203,
     RST301,
--- a/tests/tox.ini
+++ b/tests/tox.ini
@@ -1,0 +1,4 @@
+; ignore all pydocstyle errors in this folder
+
+[pydocstyle]
+add_ignore = D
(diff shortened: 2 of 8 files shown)


.pylintrc
```diff
@@ -7,7 +7,9 @@ init-import=yes
 
 [MESSAGES CONTROL]
 disable=
-    C0116,  # missing-function-docstring
+    C0114,  # done by pydocstyle
+    C0115,  # done by pydocstyle
+    C0116,  # done by pydocstyle
```
spflueger, line 12: do they conflict?

tests/tox.ini
```diff
@@ -0,0 +1,4 @@
+; ignore all pydocstyle errors in this folder
+
+[pydocstyle]
+add_ignore = D
```
spflueger, line 4: Ignoring folders can't be done in the main tox.ini? I guess thats why we had this file match workaround?

.pylintrc
redeboer: Not sure, but all documentation docstrings are handled by pydocstyle now

tests/tox.ini
redeboer: Not that I know of, see [here](http://www.pydocstyle.org/en/latest/usage.html#available-options)

redeboer: But I think this `tox.ini` in a subfolder has its advantages too

spflueger: all right

PR #93, opened 2020-06-25
ci!: set up tox test environments

### LINKED ISSUE #90: Define runtime checks in tox.ini
Since #78, developers can quickly run all style checks on the entire codebase before they push with `pre-commit`. We now need to make it easy to run all runtime checks as well in one command (`tox`, obviously). See also https://github.com/ComPWA/expertsystem/issues/45.

So we want to be able run the following with the `tox` command only:
- [x] `pytest`
- [x] `cd doc; make html`
- [x] `cd doc; make linkcheck` (requires connection!)
- [x] `cd doc; NBSPHIX_EXECUTE='' make html` (run the notebooks)
- [x] `pre-commit run -a`

redeboer:
Closes #90

### DIFF
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -4,8 +4,8 @@ jupyter
 matplotlib>=3
 nbsphinx==0.5.1
 pandas>=1
+sphinx>=2.4.4,<3.1
 sphinx-autodoc-typehints>=1.10.3
 sphinx-copybutton
-sphinx>=2.4.4,<3.1
 sphinx_rtd_theme
 sphobjinv>=2.0.1
--- a/tox.ini
+++ b/tox.ini
@@ -3,17 +3,20 @@
 ; installed this repository in 'developer mode'!
 
 [tox]
+passenv = PYTHONPATH
+skip_install = True
+skip_missing_interpreters=true
 skipsdist = True
 envlist =
-    docs,
-    pep8,
+    py,
+    doc,
+    style,
 
-[testenv:pep8]
-passenv = PYTHONPATH
-deps =
-    -rrequirements-dev.txt
+[testenv]
+whitelist_externals =
+    pytest
 commands =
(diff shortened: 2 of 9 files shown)


redeboer:
Retry DeepCode

tox.ini
```diff
@@ -3,17 +3,49 @@
...
+    make
+commands =
+    make linkcheck
+
+[testenv:style]
+whitelist_externals =
+    pre-commit
 commands =
-    flake8
+    pre-commit run -a
```
spflueger, line 48: Running style checks means running all of pre-commit?

doc/requirements.txt
```diff
@@ -4,8 +4,8 @@ jupyter
 matplotlib>=3
 nbsphinx==0.5.1
 pandas>=1
+sphinx>=2.4.4,<3.1
 sphinx-autodoc-typehints>=1.10.3
 sphinx-copybutton
-sphinx>=2.4.4,<3.1
```
spflueger, line 9: why do this dependency reorderings keep happening?

redeboer: See commit message of 6fd3ba19f3a6886c0444649ed3f80d83145408b5

tox.ini
redeboer: Yeah I start to see pre-commit as a collection of style checks (at least true if you look at the hooks that we currently implemented)

spflueger: but it also runs mypy right?

redeboer: Right, and I guess technically mypy is not a style check. But the `testenv` section names of a `tox.ini` file have a practical usage. As in, you want to keep them short and easy to type (that's why the command is the super simple '`tox`'), so that you can for instance do:

```bash
tox -e py    # run pytest
tox -e style # run 'style' checks
tox -e doc   # test-build documentation
```

PR #95, opened 2020-06-25
fix: update amplitude model recipe structure

### LINKED ISSUE #77: Modify amplitude and kinematics modules so that they can handle the new recipe file structure

redeboer:
Make the structure agree with output from
https://pypi.org/project/expertsystem/0.1.2a0

Closes #77


tests/requirements.txt
```diff
@@ -1,2 +1,3 @@
+expertsystem==0.1.2a0
```
spflueger, line 1: Why is this included here. Either we include all of the requirements or that comes from the other list

redeboer: Ah yeah this is a matter we need to discuss. Because by now we have one under the requirements for documentation (namely to run the notebooks):
https://github.com/ComPWA/tensorwaves/blob/2b85e3cd6fe4bab07b7dd958112b6e56f97f15bb/doc/requirements.txt#L2

and one under `tests` (in able to run `pytest`):
https://github.com/ComPWA/tensorwaves/blob/2b85e3cd6fe4bab07b7dd958112b6e56f97f15bb/tests/requirements.txt#L1

I did not put the expertsystem under the official requirements, because the `requirements.txt` of the `tensorwaves` package itself, because it's not a requirement for the source code itself.

For me it again show how the whole reliance on `dict` only creates the illusion that `tensorwaves` is decoupled from the `expertsystem`: in practice there is a tight coupling to the version of the `expertsystem` and it would be better to implement that as a strict dependency (even better to use recipe functionality from the expertsystem itself instead of `dict`s).

spflueger: That's true. So I guess the interface between the expert system and tensorwaves should be a graph structure instead of that dict...

examples/intensity-recipe.yaml
```diff
@@ -83,25 +83,23 @@ Intensity:
...
-                        Helicity: 0.0
-                      - Name: gamma
-                        FinalState: 2
-                        Helicity: -1
+                    - Name: f2(1270)
+                      FinalState: [3, 4]
+                      Helicity: 0.0
+                    - Name: gamma
+                      FinalState: 2
+                      Helicity: -1
```
spflueger, line 91: Is the expertsystem outputing this?

redeboer: Yeah this was requested in https://github.com/ComPWA/expertsystem/issues/73

tests/recipe/test_amplitude_creation.py
```diff
@@ -15,32 +15,42 @@ def _create_helicity_recipe():
...
+    return phsp_sample
+
+
 class TestRecipeHelicity:
-    recipe = _create_helicity_recipe()
     number_of_phsp_events = 10
+    recipe = _create_helicity_recipe()
+    particle_list = _get_particle_list(recipe)
+    kinematics = _get_kinematics(recipe)
+    phsp_sample = _generate_phsp(recipe, number_of_phsp_events)
```
spflueger: This could be a case for fixtures https://docs.pytest.org/en/stable/fixture.html

redeboer: Aha. New issue for this or fix in this PR? For me the important part in this PR is the fixes for the recipe file, whether the tests are smart pytest syntax or not would slow down this PR series

PR #106, opened 2020-07-01
feat: add benchmarks (skeleton)

### LINKED ISSUE #103: Provide benchmarks tests
Would be nice to profile/monitor this in a standardised way, so that we can see whether there are improvements upon each PR. The benchmarks should be similar in structure, probably making use of some shared façade functions.

What we probably want as input:
- Initial + final state
- Allowed intermediate states (the fewer, the faster)
- Formalism type (helicity/canonical)
- Number of:
   - phase space events
   - intensity-based events ('data')
- Stage to perform (e.g. only up to data generation)

The recipe file is generated with the expertsystem based on this input. All the rest (e.g. which amplitude generator to use), should be deduced from the recipe.

Some potential tools:
- [`pytest-benchmark`](https://pytest-benchmark.readthedocs.io/) (nicely integrated with `pytest`, though seems to be more for micro-benchmarks)
- [`pycallgraph`](https://pycallgraph.readthedocs.io/) (seems rather outdated)
- Write own timer with [`timeit`](https://docs.python.org/3/library/timeit.html)

redeboer:
Closes #103


redeboer:
@spflueger We'll need so think of a smart way to smear the initial parameters after data generation and before fit. Will also need to put the example recipe files -- created with `benchmark/create_recipe.py` -- under source control, so that we can things like omega mass width a bit more wide.

spflueger:
> @spflueger We'll need so think of a smart way to smear the initial parameters after data generation and before fit. Will also need to put the example recipe files -- created with `benchmark/create_recipe.py` -- under source control, so that we can things like omega mass width a bit more wide.

That's true. Can you open an issue for this random initial parameter problem.

I think its totally ok uploading a fixed model recipe for testing purposes.

redeboer:
> That's true. Can you open an issue for this random initial parameter problem.

There's already this one, right? https://github.com/ComPWA/expertsystem/issues/8

redeboer:
@spflueger 2d6d3001a66c1d1cd03e2ad040ed0499bc0d3ea7 still doesn't address the problem, because there's no clean way to fix or modify width and mass from the python interface.

spflueger:
> > That's true. Can you open an issue for this random initial parameter problem.
> 
> There's already this one, right? [ComPWA/expertsystem#8](https://github.com/ComPWA/expertsystem/issues/8)

This is something else. Or did I misunderstand you here? You are talking about some function which randomly picks a starting point in the optimization space right?

PR #108, opened 2020-07-07
fix: remove cast to int of helicity values

spflueger:
Note: This affects only baryonic systems in which the spins can be half
integer!


tensorwaves/physics/helicityformalism/amplitude.py
```diff
@@ -548,7 +548,7 @@ def __init__(self, name: str, helicity: int,) -> None:
     @staticmethod
     def from_dict(definition: Dict[str, Any]) -> "_HelicityParticle":
         name = str(definition["Name"])
-        helicity = int(definition["Helicity"])
+        helicity = definition["Helicity"]
```
redeboer, line 551: Does this mean there's no type hint for the arguments of `_HelicityParticle`?

redeboer:
@spflueger This PR will need to be rebased, cannot be done through github

tensorwaves/physics/helicityformalism/amplitude.py
spflueger: Not sure what you mean by this. But I think mypy can correctly deduce the return type of `from_dict`

redeboer: The interface of `_HelicityParticle` looks like
```python
class _HelicityParticle:
    def __init__(self, name: str, helicity: int,) -> None:
        self.name: str = name
        self.helicity: int = helicity
```
So you see, it expect an `int` for `helicity`, while `helicity = definition["Helicity"]` would be `Any`.

spflueger: Oh dear, you are right! Good catch, I totally missed that...

tensorwaves/physics/helicity_formalism/amplitude.py
```diff
@@ -509,9 +509,9 @@ def __call__(self, dataset: dict) -> tf.Tensor:
                 dataset[self._params.phi_name],
                 dataset[self._params.theta_name],
                 0.0,
-                2 * self._params.j,
-                2 * self._params.m,
-                2 * self._params.mprime,
+                int(2 * self._params.j),
```
redeboer, line 512: Whi is it `int` here, and in in the `_AngularProperties` it's `float`?

tensorwaves/physics/helicityformalism/amplitude.py
redeboer: But what I don't understand is why `mypy` didn't fish this out...

spflueger: Because definition is of form `Dict[str, Any]` so Any -> int is no problem. Or am I missing something?

spflueger: Or should mypy mark `Any` -> `int` is a probem?

redeboer: Ah yes that's right... So I guess the lesson is to use `Any` sparingly

redeboer: And well... these `dict`s shouldn't be there anyway, but that's the recipe structure discussion again ;)

spflueger: Yeah, `Any` prevents the checking https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-any-type

tensorwaves/physics/helicity_formalism/amplitude.py
spflueger: amplitf works with ints by multiplying everything by 2

PR #109, opened 2020-07-08
style!: fix all cspell errors

redeboer:
@spflueger This fixes the CI problems of #108 

See also https://github.com/ComPWA/expertsystem/pull/98


cspell.json
```diff
@@ -1,46 +1,115 @@
...
+        ".travis.yml",
+        ".vscode/*",
+        ".vscode/.gitignore",
+        "Makefile",
+        "codecov.yml",
         "cspell.json",
-        ".vscode/*"
+        "doc/_templates/*",
+        "doc/conf.py",
+        "examples/workflow/1_create_model.ipynb",
```
spflueger, line 28: Why ignore this file?

cspell.json
```diff
@@ -1,46 +1,115 @@
...
-        "htmlcov",
-        "ipynb",
-        "linkcheck",
+        "amplitf",
+        "atfi",
+        "cano",
+        "celltoolbar",
+        "codacy",
+        "codemirror",
+        "cparity",
```
spflueger, line 86: Should this be a word?

redeboer: Good question. I added this as one word in the `expertsystem` too, because it is used in the recipe file and is now also starting to become part of the interface of `Particle` (https://github.com/ComPWA/expertsystem/pull/100). So it comes down to a design decision.

cspell.json
redeboer: Well spotted :smirk: It's because it uses `expertsystem` v0.1.2, which still contains `helicityformalism` as one word. So as soon as we release the `expertsystem` with a new

cspell.json
spflueger: i guess we should either write `c_parity` in the code and c-parity or c parity in the docs. What do you think?

spflueger: `gparity` is in there as well btw

redeboer: Ok can do. Seems that [wiki says](https://en.wikipedia.org/wiki/C_parity) _C_ parity (no dash)

redeboer: Then I'll also fix this in https://github.com/ComPWA/expertsystem/pull/100, comes exactly at the right time.

spflueger: ok good

redeboer: a5707a164ef5dcad2f3e9543974ef0e9af2b6a7a

spflueger: Ok, since the GParity and CParity info is not used in TW this should not cause any problems.

Issue #53, opened 2020-04-30
Use binder to serve jupyter notebooks online


spflueger:
https://mybinder.org/

This would be great feature for new users to give it a try and hopefully make a good first impression

**[Estimated effort: 1d]**

redeboer:
At least one problem with mybinder seems to be that reST cells don't render.

At any rate, here are two previews:

**`tensorwaves`**
[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/ComPWA/tensorwaves/master?filepath=examples%2Fworkflow)

**`expertsystem`**
[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/ComPWA/expertsystem/master?filepath=examples%2Fjupyter%2FQuickStart.ipynb)

redeboer:
Note that the binders don't work because of the way we use the [`environment.yml`](https://github.com/ComPWA/expertsystem/blob/master/environment.yml) files. We use them only to define a virtual Conda environment in which you install `tensorwaves`/`expertsystem` in ['developer mode'](https://pwa.readthedocs.io/projects/expertsystem/en/0.1.2-alpha/install/install.html#developer-mode), i.e. just taking the source code folder as the package definition -- not installing the package.

The problem is that binder uses that `environment.yml` file as a complete definition to set up a Conda environment, meaning binder also expects you to install the package (`expertsystem`/`tensorwaves`) through that `environment.yml` definition file.

redeboer:
@spflueger If you're sure `pip install -e .` works well, we may be able to run this from the `environment.yml` file.

redeboer:
> @spflueger If you're sure pip install -e . works well, we may be able to run this from the environment.yml file.

This problem can be solved once we use the setup procedure introduced in https://github.com/ComPWA/expertsystem/pull/102

PR #110, opened 2020-07-08
build: single source of truth for dependencies

### LINKED ISSUE #107: Improve install procedure
See https://github.com/ComPWA/expertsystem/issues/103 for specification and https://github.com/ComPWA/expertsystem/pull/102 for how to address this issue. This would also be needed for #53.

redeboer:
Closes #107

See also https://github.com/ComPWA/expertsystem/pull/102


doc/requirements.txt
```diff
@@ -1,11 +1,11 @@
-doc8>=0.8.0
```
spflueger, line 1: shouldn't theses requirement files be removed now?

redeboer: Emm yes... not sure what I was doing there...

redeboer: Done: dc2ee290322fd32b36f59929706e47c66cf19c22

setup.cfg
```diff
@@ -0,0 +1,87 @@
...
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Topic :: Scientific/Engineering
+    Topic :: Scientific/Engineering :: Physics
+    Typing :: Typed
+
+[options]
+python_requires = >=3.6, <3.8
+install_requires =
+    amplitf==0.0a1
```
spflueger, line 35: we should require pip >= 19 because of tf2

environment.yml
```diff
@@ -6,4 +6,4 @@ dependencies:
   - python>=3.6,<3.8
   - pip
   - pip:
-      - -r requirements.txt
+      - -e .
```
redeboer, line 9: Maybe better use `-e .[dev]`? In that case, the install procedure for contributors using Conda would just become:
```bash
conda env create
conda activate tw
```
which previously would have been
```bash
conda env create
conda activate tw
pip install -e .[dev]
```

But note that this assumes anyone who uses Conda is also interested in contributing (i.e. the dev requirements). And this is not necessarily the case. Could especially be a problem for #53.

spflueger: hmm i would keep it optional to install the dev requirements. So you can have the run requirements be part of the dev requirements to do it all in one swoop. But im not sure how this is handled in conda

redeboer: Ok, then I leave it like this (it's also the described workflow in the install/contribute pages)

PR #114, opened 2020-07-20
docs: serve notebooks on Binder

### LINKED ISSUE #53: Use binder to serve jupyter notebooks online

redeboer:
Closes #53


redeboer:
There's a problem here: binder uses `pip install .`, while it should use `pip install .[doc]` if you want to have `pandas` installed. So we should either install `pandas` where needed -- that it, through the notebooks themselves, with something like
```python
import sys
!{sys.executable} -m pip install numpy
```
or add pandas to the package requirements. Note also that we did not fix the pandas version (it's currently `>=1`)

redeboer:
> There's a problem here: binder uses `pip install .`, while it should use `pip install .[doc]` if you want to have `pandas` installed. So we should either install `pandas` where needed -- that it, through the notebooks themselves, with something like
> 
> ```python
> import sys
> !{sys.executable} -m pip install pandas>=1
> ```
> 
> or add pandas to the package requirements. Note also that we did not fix the pandas version (it's currently `>=1`)

@Leongrim this has been fixed by 9bab91e84b0b12b69a478c134ec60569b19770a0

PR #116, opened 2020-07-23
build: upgrade to expertsystem 0.2.0

redeboer:
Version 0.2.0 simplifies usage of the particle database and makes writing the recipe file easier. The workflow notebooks therefore also become a bit easier to use.

Additional modifications:
* Added a note on how to update (that is, both `git pull` and `pip install -e .`)
* Intersphinx is used not only used for cross-references to external APIs, but also to their documentation pages (e.g. the quickstart of the `expertsystem`).
* `conf.py` now follows the same structure as that of the `expertsystem`.


doc/install.rst
```diff
@@ -165,3 +152,25 @@ For more thorough testing, navigate back to the you can run the unit tests:
...
+
+When new commits are merged into the `master branch
+<https://github.com/ComPWA/tensorwaves/tree/master>`_, you need to update your
+:ref:`local copy of the source code <local-repository>` as follows:
+
+.. code-block:: shell
+
+  git checkout master
+  git pull
+  pip install -e .
```
spflueger, line 167: I believe the pip install -e is not even necessary if you previously installed it in editable mode

redeboer: Yeah but see what I [wrote underneath](https://github.com/ComPWA/tensorwaves/pull/116/files#diff-b50754a7065a1dece18f634c6b5b2633R171-R172): it's for potential changes in the dependencies

spflueger: I see

Issue #118, opened 2020-08-13
Upgrade expertsystem


redeboer:
Reason:
* More conventient handling of particle info
* @Leongrim has some bug

Ideally, we wait until the `expertsystem` is further refactored and has a more stable release, but this is needed for the ongoing trial analyses.

redeboer:
Better to wait until https://github.com/ComPWA/expertsystem/milestone/13

PR #121, opened 2020-08-18
feat: add default dynamics RelativisticBW

redeboer:
Addresses #118, but does not close it. Since #116, dynamics are not provided from the recipe file, so we need some default.

### DIFF
--- a/tensorwaves/physics/helicity_formalism/amplitude.py
+++ b/tensorwaves/physics/helicity_formalism/amplitude.py
@@ -40,6 +40,9 @@ from ._recipe_tools import extract_value
 from .kinematics import HelicityKinematics, SubSystem
 
 
+_DEFAULT_DYNAMICS = "RelativisticBreitWigner"
+
+
 class IntensityTF(Function):
     """Implementation of the `~.Function` interface using tensorflow.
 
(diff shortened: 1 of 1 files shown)


redeboer:
> if we want the same default behavior as before

For me this PR is just putting the finger in the dike: a bug fix that is not meant to tackle the bigger design.

spflueger:
> > if we want the same default behavior as before
> 
> For me this PR is just putting the finger in the dike: a bug fix that is not meant to tackle the bigger design.

But like I said this version will not deliver the same results as before, since no the top node is also using a BW unless the user specifies that explicitly

redeboer:
Yeah but that is already the case since #116: the program just dies because when it searches for non-existing dynamics.

spflueger:
Ok, I was missing the info that non dynamics is still specified by the expert system on the top node by default. So the default behavior does stay the same. Not the prettiest bugfix though.

redeboer:
Yes, so to my understanding, the expertsystem already adds 'non-dynamic' dynamics for the top node:
https://github.com/ComPWA/expertsystem/blob/94231dea8b200097edd1a8c8be1d56a5210a0735/expertsystem/amplitude/_yaml_adapter.py#L31-L36

It's super dirty, because this happens only for YAML, but we don't really want to support XML for `tensorwaves` anyway.

PR #126, opened 2020-08-19
ci!: equalize dev set-up

### LINKED ISSUE #119: Import new developer configuration of expertsystem
- [x] Migrate to GitHub Actions
- [x] PR linting
- [x] Badges
- [x] Branch coverage
- [x] `markdownlint`
- [x] Move `mypy` config to `setup.cfg` so you can just run `mypy .`
- [ ] ~`data-science-types` for typing of `numpy`, `pandas`, and `matplotlib`~ --> #125

redeboer:
Closes #119 

This PR equalizes the developer setup of `tensorwaves` with that of the `expertsystem`.

## Summary
- ci: enforce `markdownlint`
- ci: update `pre-commit`
- ci: move `mypy` configuration to `setup.cfg`
- ci: set max line width in Prettier formatted files to 79
- ci: allow warnings on RTD (already checked in GitHub Actions)
- ci: add VSCode Coverage Gutters extension configuration
- ci: specify minimal patch coverage
- ci: flag more `cspell` words
- ci: add Jupyter Lab (+extensions) to dev requirements
- chore: clean up `.gitignore` files (e.g. add excludes, which is handy for file searches in IDEs that hides files under `.gitignore`)
- docs: remove `sphinx_autodoc_typehints` in favour of newer version of Sphinx
- docs: import improvements to the Contribute page and the Install page from the `expertsystem`
- docs: add `myst-parser` for Markdown support in the Sphinx pages
- fix: run `cspell` over all Jupyter notebooks


tox.ini
```diff
@@ -16,7 +12,7 @@ envlist =
 whitelist_externals =
     pytest
 commands =
-    pytest tests {posargs} \
+    pytest {posargs:tests} \
```
spflueger, line 15: what the use of posargs?

tox.ini
```diff
@@ -47,13 +43,17 @@ commands =
     make linkcheck
 
 [testenv:sty]
+setenv =
+    SKIP = mypy
 whitelist_externals =
+    mypy
     pre-commit
 commands =
+    mypy .  # run separately because of potential caching problems
```
spflueger, line 52: which caching problems?

labels.toml
```diff
@@ -0,0 +1,49 @@
+[Bug]
```
spflueger, line 1: Are those issue labels used by zenhub?

tox.ini
redeboer: This allows you to pass on arguments when running `tox`, e.g.
```bash
tox -e py tests/physics
```
Here, `tests/physics` is the `posargs`.

tox.ini
redeboer: Sometimes `mypy` messes up if you run it with `pre-commit run mypy -a` (I think because that would run over all files under version control).

labels.toml
redeboer: Actually it's just a way of putting the GitHub issue labels under version control. It's not automated (yet), all I did was
```bash
labels fetch -o ComPWA -r tensorwaves
```
plus some token settings etc. (which is the hard part to automate). For more info see [`labels`](https://pypi.org/project/labels/).

tox.ini
spflueger: I see, so its either posargs or if not given then tests by default?

tox.ini
spflueger: so what is different to calling `mypy .` then?

tox.ini
redeboer: Yep, it's the same as in bash scripts or functions (e.g. `${0:-default argument}`)

labels.toml
spflueger: Ok

tox.ini
spflueger: Ok, did not know that. Thanks

tox.ini
redeboer: `mypy .` just runs over the files it can find according to these rules
https://github.com/ComPWA/tensorwaves/blob/6ccb5c8ea4f29509c8e330742a381d7c92a230f7/setup.cfg#L90-L97

The `pre-commit` hook also has some sort of filter like that
https://github.com/ComPWA/tensorwaves/blob/6ccb5c8ea4f29509c8e330742a381d7c92a230f7/.pre-commit-config.yaml#L53-L58

but somehow it doesn't work well. This solution worked well so far in the `expertsystem`.

spflueger: so in `pre-commit` those can be used but dont work in the same way?

redeboer: `pre-commit` for `mypy` works if it's just running over staged files. Just _can_ run into problems when running over all files (`run -a`).

redeboer: As a side note: I have the impression that `pre-commit run mypy -a` fails when the interfaces change. Then, if you run that command again, the problem disappears. In contrast, `mypy .` never causes problems.

spflueger: Hmm ok.

spflueger (approved):

PR #113, opened 2020-07-20
build: upgrade expertsystem

redeboer:

PR #128, opened 2020-08-28
build!: upgrade to TF2.3

### LINKED ISSUE #127: Upgrade to TF2.3
Python 3.8 is supported since TF2.2

redeboer:
Closes #127

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python>=3.6,<3.8
+  - python>=3.6
   - pip>=19
   - pip:
       - -e .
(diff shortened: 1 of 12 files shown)


environment.yml
```diff
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - python>=3.6,<3.8
+  - python>=3.6
```
spflueger, line 6: Does 3.6 work? Since the expertsystem requires > 3.6

redeboer: No, but that will be tackled by #113. In this PR, we still use `expertsystem==0.2.0`

spflueger (approved):

PR #137, opened 2020-10-19
build: upgrade to expertsystem 0.5.1

redeboer:


examples/workflow/2_generate_data.ipynb
```diff
@@ -69,7 +69,7 @@
     "kin = HelicityKinematics.from_model(model)\n",
     "print(\"Initial state mass:\", kin.reaction_kinematics_info.initial_state_masses)\n",
     "print(\"Final state masses:\", kin.reaction_kinematics_info.final_state_masses)\n",
-    "print(\"Involved particles:\", [p.name for p in model.particles])"
+    "print(\"Involved particles:\", model.particles.names)"
```
spflueger, line 72: is this names property needed then?

redeboer: It's more efficient: it just returns the keys of the underlying dictionary. But in the first place it's just more conventient

spflueger: Oh sorry, I somehow misread this change the opposite way (that it changed from names to the list comprehension)....

spflueger (approved):

Issue #122, opened 2020-08-18
Missing invariant masses for the kinematics


Leongrim:
At the moment it seems like the kinematics only creates the invariant masse for subsystems containing resonances because this are the only ones that are needed for the fitting.

spflueger:
Will will have to see who to assign this task to. Any volunteers? 

The idea would be to generate all different subsystems using https://docs.python.org/3.7/library/itertools.html#itertools.combinations . Then simply register those Subsystems in the Kinematics instance.

redeboer:
> Any volunteers?

:hand: :unamused: 
But probably best to do this together with @Leongrim because he can test it on a use case.

spflueger:
For the record, this is how it was done in ComPWA (c++) https://github.com/ComPWA/ComPWA/blob/master/Physics/HelicityFormalism/HelicityKinematics.cpp#L296

PR #142, opened 2020-10-27
docs: improve documentation with plots

redeboer:
Summary of changes:
* Removed text about downloading the notebook, because this is already proivded by the Sphinx Book theme.
* Explained a bit more about the example decay channel
* Use all _f0_ resonances in the example decay channel
* Show how to convert the state IDs to particle names
* Illustrate how to plot the fit result with `matplotlib`
* Added a snippet that shows how to indicate the resonance positions in the plots
* Added a snippet for exporting the fit result (but it's not yet used to import it back)


redeboer:
@Leongrim @sebastianJaeger These changes additions are to help you guys, could you have a look?

@spflueger You can have a look as well, but I did not modify the source code.

redeboer:
So for the record: best to review the rendered pages
https://pwa--142.org.readthedocs.build/projects/tensorwaves/en/142/usage.html

You can also download the notebooks from there (upper right) and run them yourself to see if you would like to add some more code snippets that are interesting to the physicist. (Better would of course be to checkout the branch of this PR.)

Leongrim:
In <a href="https://pwa--142.org.readthedocs.build/projects/tensorwaves/en/142/usage/2_generate_data.html">Step 2</a> it is mentioned that not all the invariant masses are registered. Are all the angles registered?

redeboer:
Good point. Actually it's just a quick remark that had to be there while #122 is still out there.

Unfortunately, the PR preview does not render output, but from the `master` branch, I see [this](https://pwa.readthedocs.io/projects/tensorwaves/en/latest/usage/2_generate_data.html#2.4-Visualize-kinematic-variables):

```python
>>> data_set = kin.convert(data_sample)
>>> list(data_set.keys())
['mSq_2_3_4',
 'mSq_3_4',
 'mSq_2',
 'mSq_3',
 'mSq_4',
 'theta+3_4+2',
 'phi+3_4+2',
 'theta+3+4_vs_2',
 'phi+3+4_vs_2']
```

Meaning also only the angles of the relevant subsystems are generated. The best solution would be to create some `register_all_subsystem` method (to solve #122), just like ComPWA's [`createAllSubsystems`](https://compwa.github.io/ComPWA/classComPWA_1_1Physics_1_1HelicityFormalism_1_1HelicityKinematics.html#a9e6dca848564a15ff65f1af0fe8383b4). It's better to offer that as a method (not the default behaviour), because having all subsystems generated by default would create large data sets.

Leongrim (approved):

PR #148, opened 2020-11-10
test: add test for generate_data

redeboer:
Needed for #146

### DIFF
--- a/tests/data/test_generate.py
+++ b/tests/data/test_generate.py
@@ -1,6 +1,9 @@
+import expertsystem.amplitude.model as es
+import numpy as np
 import pytest
 
-from tensorwaves.data.generate import generate_phsp
+from tensorwaves.data.generate import generate_data, generate_phsp
+from tensorwaves.physics.helicity_formalism.amplitude import IntensityBuilder
 from tensorwaves.physics.helicity_formalism.kinematics import (
     HelicityKinematics,
     ParticleReactionKinematicsInfo,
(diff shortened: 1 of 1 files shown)


tests/data/test_generate.py
```diff
@@ -41,3 +44,41 @@ def test_shape_generate_phsp(
...
+        "mSq_3_4",
+        "mSq_4",
+        "phi+3+4_vs_2",
+        "phi+3_4+2",
+        "theta+3+4_vs_2",
+        "theta+3_4+2",
+    }
+    assert pytest.approx(data_set["mSq_2"].mean()) == 0
+    assert pytest.approx(data_set["mSq_3"].mean()) == data_set["mSq_4"].mean()
+    assert pytest.approx(data_set["mSq_3_4"].mean(), abs=1e-1) == 1
```
spflueger, line 84: I would test these kinematic related things like masses inside the phase space generation, since that is related to that. The generate data is related to the model, so we should perform checks that test for some expected distribution or so.

redeboer: Which checks do you suggest? (The masses are already checked in `test_generate_phsp`.)

spflueger: There are some distribution checks in pycompwa that we could move here. But they are slighly more complicated though and took a few more seconds to run.

For now we could just check that the sample sizes match and extend the tests to distribution verifications later on

redeboer: ff8f254

spflueger (approved):

Issue #99, opened 2020-06-26
Add functionality to export and import fit result


redeboer:
[Step 3](https://pwa.readthedocs.io/projects/tensorwaves/en/latest/usage/3_perform_fit.html) of the [workflow notebooks](https://pwa.readthedocs.io/projects/tensorwaves/en/latest/usage.html) is incomplete: it does not export the fit result, as we have that functionality. Practically, this will be necessary for Step 4 (compare the [corresponding step in the ComPWA Workflow](https://compwa.github.io/usage/workflow.html)).

redeboer:
In the latest version, we point out that one can simply dump the `dict` of fit parameters with values to JSON
https://pwa.readthedocs.io/projects/tensorwaves/en/0.1.2/usage/3_perform_fit.html#export-fit-result

That would already be a good starting point: perhaps write some wrapper for the procedure described there, then call that wrapper at some natural recurring point in the fitting process. That will be the harder part, I guess, to define a point where the optimizer writes its current fit parameters to disk.

Next, we could also think about some interfaces to import the JSON file and use those values as 'initial' parameters.

PR #154, opened 2020-11-11
feat: add intermediate result YAML saving

### LINKED ISSUE #158: Provide procedure/interface to interrupt fit and pick up latest fit result

sebastianJaeger:
I added a functionality that regularly saves intermediate fit results, so not all of the progress is lost upon abortion of the fit or a death of the process.

Closes #158


src/tensorwaves/optimizer/minuit.py
```diff
@@ -69,3 +81,18 @@ def __func(pars: list) -> float:
...
         results["time"] = end_time - start_time
         return results
+
+
+def dump_parameters(parameters: dict, outputfile: str) -> None:
+    output = {
+        "Parameters": {
+            name: float(value) for name, value in parameters.items()
+        }
+    }
```
spflueger, line 91: I would probably just dump the dictionary as it is, without wrapping it in the `"Parameters"` (That's a bit of a preference though). I probably would put parameters in the automated temporary filename. Maybe @redeboer has an opinion about this.

src/tensorwaves/optimizer/minuit.py
```diff
@@ -69,3 +81,18 @@ def __func(pars: list) -> float:
...
+    with open(outputfile, "w") as yaml_file:
+        yaml.dump(
+            output,
+            yaml_file,
+            sort_keys=False,
+            default_flow_style=False,
+        )
```
spflueger, line 98: Just an idea (that feature does not have to also be implemented here): We could think about a feature that appends new Parameters to have really a path of how the Optimizer traveled.

spflueger (commented):
@sebastianJaeger Thanks for the feature! I added a few comments.

One general thing is, that it is rather hard to test this functionality nicely. But that's more of a design issue, and I would propose to implement this functionality in a decorator (the decorator can be tested with a mock quite nicely). 

If you do now have fun doing this, I can understand. But this would be a perfect time to test our new branching model in action. The idea is to push this code into a new "feature" branch. In this branch the "code quality" are not as strict as in the master branch, but allow quick additions of new features. Then those branches can be refactored and merge with the master later on. @redeboer can tell a bit more about the details

src/tensorwaves/optimizer/minuit.py
```diff
@@ -14,13 +17,17 @@ class Minuit2(Optimizer):
...
+        self, filename: Optional[str] = None, n_iterations: int = 10
+    ) -> None:
+        self.__filename = filename
+        self.__n_iterations = n_iterations
 
     def optimize(self, estimator: Estimator, initial_parameters: dict) -> dict:
         parameters = initial_parameters
 
         function_calls = 0
+        progress_bar = tqdm()
```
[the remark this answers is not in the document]
sebastianJaeger: It does not get to know when its finished, but at least you get to see how fast you run, sicne it shows the number of iterations and the time per iteration, so you have a rough guess how quickly the computation goes. In my opinion thats still better than no current status output at all.

src/tensorwaves/optimizer/minuit.py
sebastianJaeger: That might me usefull, but I dont think that should be enabled by default, since it probably complicates the original task this was implemented for.

redeboer:
Here are a few thoughts:
- We should try to follow the interfaces of [keras optimizers](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/Optimizer)
- If we write to a format that can be understood by [tensorboard](https://www.tensorflow.org/tensorboard/scalars_and_keras), we may be able to use `tensorboard` to monitor fit progress.
- For that we need to write to different sequential log files, see also https://github.com/ComPWA/tensorwaves/pull/154#discussion_r521906639

> This would be a perfect time to test our new branching model in action. The idea is to push this code into a new "feature" branch. In this branch the "code quality" are not as strict as in the master branch, but allow quick additions of new features. Then those branches can be refactored and merge with the master later on. @redeboer can tell a bit more about the details

Since this concerns quite a few issues, I created a branch [99-export-fit](https://github.com/ComPWA/tensorwaves/tree/epic/99-export-fit) that is to address #99.

src/tensorwaves/optimizer/minuit.py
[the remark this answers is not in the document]
redeboer: Update: #157 implements a `ProgressBar` callback that prints the current estimator value in addition to the time progressed. We can also print other stats that way.

src/tensorwaves/optimizer/minuit.py
redeboer: This could be handled by creating a new `Callback` implementation (#164) and registering values in, say, a `pandas.DataFrame`.

Side note: it would be nice to have the fit running in the background (`asyncio`?), so that you can just `Minuit.optimize` in a notebook and monitor (dynamically -- not by continuously reading back a file) the fit parameters in that `DataFrame`.

redeboer:
I updated this PR a bit with the new callback syntax (#164). Now it should also close #158, as illustrated in the notebook.

@sebastianJaeger, have a look here for the new implementation
https://github.com/ComPWA/tensorwaves/pull/154/commits/2d621a560ad7b6d13fbf33e4c2c61632c63b7a54?file-filters%5B%5D=.py#diff-9f533f5231a0e5fcd84806765d6bbf0282451172465fee983ac5df640128cceeR34-R77

src/tensorwaves/optimizer/minuit.py
redeboer: I still wrapped the parameters in a subsection, because there's also a `"Time"` section and `"EstimatorValue"`.

redeboer (approved):

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -30,6 +31,50 @@ def finalize(self) -> None:
...
+
+        .. code-block:: bash
+
+            tensorboard --logdir logs
+        """
+        self.__iteration = 0
+        self.__step_size = step_size
+        self.__stream = open(filename, "w")
+
+    def __call__(self, parameters: dict, estimator_value: float) -> None:
```
spflueger, line 53: The Callbacks are actually called on every function evaluation... Unfortunately we don't have access to the iterations (at least when using minuit).

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -30,6 +31,50 @@ def finalize(self) -> None:
...
+        self.__iteration += 1
+        if self.__iteration % self.__step_size != 0:
+            return
+        output_dict = {
+            "Time": datetime.now(),
+            "Iteration": self.__iteration,
+            "EstimatorValue": float(estimator_value),
```
spflueger, line 60: It would be good if we would be able to pass the estimator label through as well. Maybe through the init?

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -30,6 +31,50 @@ def finalize(self) -> None:
...
+            "Parameters": {
+                name: float(value) for name, value in parameters.items()
+            },
+        }
+        _empty_file(self.__stream)
```
spflueger, line 65: we dont want to kept appending?

spflueger (approved):
All in all looks good. Just some small remarks that are not urgent (since its only going into the epic branch).

src/tensorwaves/optimizer/callbacks.py
redeboer: No, that would conflict with YAML syntax

src/tensorwaves/optimizer/callbacks.py
redeboer: Yeah, that was one of the reasons for posting #165

src/tensorwaves/optimizer/callbacks.py
redeboer: Have a look 171b6e5 (it can be reverted)

src/tensorwaves/optimizer/callbacks.py
spflueger: Not if everything is wrapped into an array right? But ok that can be a future improvement

redeboer: Yeah but then YAML is perhaps not the appropriate format (rather csv or so)

spflueger: Just another idea. Maybe Yaml is also not the best choice here. And the structure. If we would append there is like a 70% overhead in terms of storage, since the parameter names would be written to disk every time. Some table structure with a header would be more useful.

src/tensorwaves/optimizer/callbacks.py
spflueger: Looks ok. But would be a bit inconvenient for appending  see https://github.com/ComPWA/tensorwaves/pull/154#discussion_r525941074

src/tensorwaves/optimizer/callbacks.py
spflueger: I meant more that `iterations` should be renamed to `function_calls`. As iteration is the incorrect word here

spflueger:
I added some more ideas. But depending on how fast this functionality is needed it can be merged to the epic branch and improved there later on. Hence i approved the PR

PR #155, opened 2020-11-12
feat: add variable logging functionality using tf

spflueger:
Adds a tf file logging decorator that can be used in Optimizers to add log variables. The decorator wraps a model updating function.

Note its some preliminary code that shows what I had in mind with the decorator. You can see it in action by calling tensorboard like this
```sh
tensorboard --logdir logs
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.npy
 *.pdf
 *.png
+*.v2
 *.xml
 *.yaml
 *.yml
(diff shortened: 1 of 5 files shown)


redeboer:
I added a small note on TensorBoard in the fitting notebook (3badb7) to see whether/how it renders on RTD (7f4dd01). Two remaining questions:
- How to pick up a fit from these log files?
- How to export to human-readable format (unless we're fine with having to launch tensorboard to monitor the status).

redeboer:
Okay, update: directly rendering tensorboard on RTD wasn't a smart idea
https://pwa--155.org.readthedocs.build/projects/tensorwaves/en/155/usage/3_perform_fit.html#optimize-the-intensity-model
but still the two cells should be useful if not executed.

PR #159, opened 2020-11-13
epic: export and import fit result

### LINKED ISSUE #99: Add functionality to export and import fit result

redeboer:
Closes #99


redeboer:
I set this PR to ready for review, although #165 is still in the epic. We have to decide how urgent that issue is and whether it's the right way to go.

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -0,0 +1,158 @@
...
+        """Log fit parameters and the estimator value to a `tf.summary`.
+
+        The logs can be viewed with `TensorBoard
+        <https://www.tensorflow.org/tensorboard>`_ via:
+
+        .. code-block:: bash
+
+            tensorboard --logdir logs
+        """
+        self.__iteration = 0
```
spflueger, line 52: `iteration` should be renamed to `function_calls`. Not just in this class. This is just related to the fact that the callbacks in Minuit are called on every function call made. Iteration info could be available in some other optimizers. To be able to handle both (since minuit does not provide this), some design changes would be needed.

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -0,0 +1,158 @@
...
+            },
+        }
+        _empty_file(self.__stream)
+        yaml.dump(
+            output_dict,
+            self.__stream,
+            sort_keys=False,
+            Dumper=_IncreasedIndent,
+            default_flow_style=False,
+        )
```
spflueger, line 81: As mentioned in #154, I think it would be more useful to have a history of the fit result instead of just writing the latest state. But there is probably some preference of mine in it. Anyways, if we leave it like this (saving only the latest result ), then it could be useful to mention that in the docstring

spflueger (changes requested):

src/tensorwaves/optimizer/callbacks.py
redeboer: See also #166

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def __call__(self, parameters: dict, estimator_value: float) -> None:
+        self.__progress_bar.set_postfix({"estimator": estimator_value})
+        self.__progress_bar.update()
+
+    def finalize(self) -> None:
+        self.__progress_bar.close()
+
+
+class YAMLSummary:
```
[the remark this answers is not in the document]
redeboer: Good point, I put it under #166

src/tensorwaves/optimizer/callbacks.py
redeboer: 0a34685

src/tensorwaves/optimizer/callbacks.py
spflueger: Sounds good, but the current two summaries are covering most of the needs for now already, so its not that urgent. I'm not requesting this change. Just wanted to point out. Though we should mention that this yaml summary is a bit different than the standard loggers, which use appending.

redeboer:
Notice that the build on RTD fails if the notebooks are run
https://readthedocs.org/projects/tensorwaves/builds/12372875

This could be because the fit is now run twice (albeit with fewer free parameters), although it doesn't seem like the runtime increased that significantly. Compare e.g. the CI for #159
https://github.com/ComPWA/tensorwaves/pull/159/checks?check_run_id=1423128675
with the one for this PR
https://github.com/ComPWA/tensorwaves/runs/1419960001?check_suite_focus=true

Both are around 3min

src/tensorwaves/optimizer/callbacks.py
redeboer: #159 is read for review

tests/conftest.py
```diff
@@ -24,11 +50,82 @@ def canonical_model() -> AmplitudeModel:
...
+) -> dict:
+    optimizer = Minuit2(
+        callback=CallbackList(
+            [
+                CSVSummary(filename=output_dir + "fit_traceback.csv"),
+                YAMLSummary(filename=output_dir + "fit_result.yml"),
+            ]
+        )
+    )
+    return optimizer.optimize(estimator, free_parameters)
```
spflueger, line 118: Generally I don't like to have that many fixtures connected to one specific scenario, as this couples all of the tests together somewhat. For testing its ideal to also keep things decoupled so that we can quickly see which parts are going wrong. Concretely that would mean, that I would mock an Optimizer to test the callbacks. Once that is changed  the `free_parameters` fixture becomes redundant (can go directly to the Minuit2 test).

Generally we also have to start adding more edge cases to our tests, like an empty list of initial_parameters in the fit. That we expect some Error etc. Or a filename that already exists or an invalid path etc in the Callback tests.

spflueger (approved):
Bood job. A general remark about the tests. We have to make sure we test things a bit more thoroughly, especially also edge cases. The test we have currently just create an artificially high test coverage, which I find quite dangerous.

So we have to make some improvements on that regard. That can go into different PRs though. Possibly also good issues for @Leongrim and @sebastianJaeger to get acquaint with.

tests/conftest.py
redeboer: Better handled through #103

PR #164, opened 2020-11-16
refactor: log optimize process via callbacks

### LINKED ISSUE #157: Enable optimize logging via callbacks
Follow-up to #155 

Some thoughts:
- Find out how to convert the `tf.Summary` produced by 
  https://github.com/ComPWA/tensorwaves/blob/031a8002464e57be4ac24a96d82aaad372d21f79/src/tensorwaves/optimizer/logging.py#L19-L54

  to human-readable format (YAML). Can probably be done by converting the logs to `pandas.DataFrame`, see instructions [here](https://www.tensorflow.org/tensorboard/dataframe_api).

- Provide different ways to export the intermediate status of a fit process (see e.g. proposed #154). Preferred implementation would be that of [`callbacks`](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks) that can be passed to the `optimize` call. See workflow [here](https://www.tensorflow.org/tensorboard/scalars_and_keras).

redeboer:
Closes #157


redeboer:
@spflueger Have a look what you think of this setup. I'm not really satisfied with it, but it illustrates the idea.

As a next step, an additional callback writer could be implemented to address https://github.com/ComPWA/tensorwaves/issues/157#issuecomment-727812520, and then #158

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -0,0 +1,92 @@
...
+from typing import Any, Iterable, List, Optional
+
+import tensorflow as tf
+from tqdm import tqdm
+
+
+class Callback(ABC):
+    @abstractmethod
+    def __init__(self, **kwargs: Any) -> None:
+        pass
```
spflueger, line 14: I don't like `__init__` to be part of the interface. Since that would mean the arguments (which would come from the optimizer) have to be passed at the construction of a concrete `Callback`. So I would rather change that into a normal function. Maybe like `def initialize(kwargs)`

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @abstractmethod
+    def __call__(self, **kwargs: Any) -> None:
```
spflueger, line 17: I don't like kwargs in an interface, as it blurs the whole interface. Whats the reasoning for that? I mean there is only as rather specific set of information that can be passed from the optimizer to the callback (currently we only can think of parameters and the estimator value)

src/tensorwaves/optimizer/minuit.py
```diff
@@ -16,27 +16,30 @@ class Minuit2(Optimizer):
...
+            estimator_value = estimator()
+            self.__callbacks.call_all(
+                parameters=params,
+                estimator_value=estimator_value,
+            )
+            return estimator_value
 
-        def __func(pars: list) -> float:
-            """Wrap the estimator."""
+        def __wrapped_function(pars: list) -> float:
```
spflueger, line 36: I originally split out this `__call_estimator` function to work with this decorator pattern. Since we are using the callback, I would merge `__call_estimator` with `__wrapped_function` again.

src/tensorwaves/optimizer/minuit.py
```diff
@@ -16,27 +16,30 @@ class Minuit2(Optimizer):
     Implements the `~.interfaces.Optimizer` interface.
     """
 
-    def __init__(self) -> None:
-        pass
+    def __init__(self, callbacks: Optional[Iterable[Callback]] = None) -> None:
```
spflueger, line 19: I like this `CallbackList` idea. Can't the callbacks argument here be reduced to `Optional[Callback]` and in case multiple callbacks are wanted, a `CallbackList` can be used as the argument. For that to work the `CallbackList` should implement the `Callback` interface though.

spflueger (commented):
Looks quite good! I have a design related comments

src/tensorwaves/optimizer/callbacks.py
redeboer: Oh yes, actually this is a remnant of when I still used `Type[Callback]` in the `CallbackList`, but that turned into a mess. So as it is now, this `__init__` interface can be removed without problems.

src/tensorwaves/optimizer/callbacks.py
redeboer: I also don't really like `kwargs` and initially had `parameters` and `estimator_value` in the interface. But then I had to use those in the `ProgressBar` as well without them being used, so then I wondered whether it's a good idea to enforce these arguments. Wouldn't such an interface be too tailored to how it is  used in the current `Optimizer` implementation (currently only `Minuit2`)?

As for using `kwargs` at all: I would say you would need that if you want to register additional information. So for me the question is more whether or not to enforce `parameters` and `estimator_value`.

For now I see a `Callback` as a more abstract procedure: a hook that can be called and that can be finalised.

src/tensorwaves/optimizer/minuit.py
redeboer: The idea comes from
https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/CallbackList
though the existing implementation is just to clean up the code in the `optimize` method a bit.

src/tensorwaves/optimizer/callbacks.py
redeboer: > I mean there is only as rather specific set of information that can be passed from the optimizer to the callback (currently we only can think of parameters and the estimator value)

An argument for this could be that the `callbacks` module is located under `optimizer`

src/tensorwaves/optimizer/callbacks.py
redeboer: 5aa1b29

src/tensorwaves/optimizer/minuit.py
redeboer: 74bda11

src/tensorwaves/optimizer/minuit.py
redeboer: Looks better now :smirk: 5dba9c4

src/tensorwaves/optimizer/callbacks.py
spflueger: > I also don't really like `kwargs` and initially had `parameters` and `estimator_value` in the interface. But then I had to use those in the `ProgressBar` as well without them being used, so then I wondered whether it's a good idea to enforce these arguments. Wouldn't such an interface be too tailored to how it is used in the current `Optimizer` implementation (currently only `Minuit2`)?
> 
That is true. But that's the whole point of an interface. That is standardizes the procedure and allow to "hot swap" modules and things will continue to work. In other words this kwargs "interface" violates the Liskov Substitution Principle (btw did you get any error message on that?). As an example, lets say one Callback requires some argument `estimator_value`, another Callback maybe `log_likelihood` (or whatever). But how do you know in the Optimizer, which Callbacks need which arguments? You can't unless you define an interface with a concrete specification.

> As for using `kwargs` at all: I would say you would need that if you want to register additional information. So for me the question is more whether or not to enforce `parameters` and `estimator_value`.
> 
> For now I see a `Callback` as a more abstract procedure: a hook that can be called and that can be finalised.

If we need to be more flexible for more custom values. Then maybe the access to that info has to be given "by reference" beforehand. So that the callback receives the estimator, parameters and maybe some part of the optimizer and can extract all of the relevant info from there. Btw some info that might be interesting to be logged in the future would be the gradient info (if available)

src/tensorwaves/optimizer/minuit.py
spflueger: Yeah, but they also make the `CallbackList` implement the `Callback`  interface right?

src/tensorwaves/optimizer/callbacks.py
redeboer: Good points :+1: Here are two versions that may fix it:
15b1dbe (enforce `parameters` and `estimator_value`)
b69e549 (remove kwargs as well).

redeboer: > Then maybe the access to that info has to be given "by reference" beforehand.

How about defining a `FitResult` class? (Would be a follow-up issue)

src/tensorwaves/optimizer/minuit.py
```diff
@@ -16,27 +16,23 @@ class Minuit2(Optimizer):
...
-
-        def __func(pars: list) -> float:
-            """Wrap the estimator."""
+        def __wrapped_function(pars: list) -> float:
             for i, k in enumerate(parameters.keys()):
                 parameters[k] = pars[i]
-            return __call_estimator(parameters)
+            estimator.update_parameters(parameters)
+            estimator_value = estimator()
+            if self.__callback is not None:
```
spflueger, line 30: I just realized that this if case at that point is a bit smelly. I'm not sure how smart python is in optimizing this. But we could improve this by checking at the init level if callbacks are given and if so we call a second version of a `__wrapped_function` which includes the callback.

redeboer: Alternative: set an empty `CallbackList` in case of `None` in `Minuit2.__init__`

src/tensorwaves/optimizer/callbacks.py
spflueger: > Good points  Here are two versions that may fix it:
> [15b1dbe](https://github.com/ComPWA/tensorwaves/commit/15b1dbed73e81c02b66e332a11688c93cbf28abc) (enforce `parameters` and `estimator_value`)
> [b69e549](https://github.com/ComPWA/tensorwaves/commit/b69e549d7c422c6bb2418a8e005ef414746d55c2) (remove kwargs as well).

Looks good. We will have to see if we can keep this interface stable. If it requires to much customized behaviors of Optimizers to loggers, then we have to reconsider this design...

spflueger: > > Then maybe the access to that info has to be given "by reference" beforehand.
> 
> How about defining a `FitResult` class? (Would be a follow-up issue)

In ComPWA (c++ codebase) this FitResult was templated. Because this output is strongly dependent on/customized by the Optimizer that is used.

spflueger (approved):
I have one last performance related comment, but otherwise I have nothing more to add. We have to see if this design is stable enough for us. That `__call__(parameters, estimator_value)` is still a bit troubling...
Tricky topic. Anyways, good job!

src/tensorwaves/optimizer/minuit.py
redeboer: 7f02297

redeboer:
Reverted 6f80184 because `if` statements indeed seem to be faster (credit: @spflueger)

PR #168, opened 2020-11-18
feat!: generate deterministic data samples

### LINKED ISSUE #59: Deterministic behavior for phasespace generator

redeboer:
Closes #59


tests/data/test_generate.py
```diff
@@ -1,43 +1,139 @@
...
+    model = helicity_model
+    kinematics = HelicityKinematics.from_model(model)
+    phsp_sample = generate_phsp(100, kinematics)
+    builder = IntensityBuilder(model.particles, kinematics, phsp_sample)
+    intensity = builder.create_intensity(model)
+    sample_size = 3
+    rng = TFUniformRealNumberGenerator(seed=0)
+    sample = generate_data(
+        sample_size, kinematics, intensity, random_generator=rng
+    )
```
redeboer, line 23: So, what happens here is a bit tricky, as rightly pointed out https://github.com/ComPWA/tensorwaves/issues/59#issuecomment-716658526: There's just one seed, and one generator that's used by both `generate_data` and `generate_phsp`.

src/tensorwaves/data/tf_phasespace.py
```diff
@@ -25,21 +28,31 @@ def __init__(
         )
 
     def generate(
-        self, size: int, random_generator: UniformRealNumberGenerator
+        self,
+        size: int,
+        seed_or_rng: Optional[Union[UniformRealNumberGenerator, int]] = None,
```
redeboer, line 33: Naming may be approved: currently sounds more like a boolean.

spflueger: I believe the pythonic thing would be to split it up into two arguments, one seed one rng, and rng takes priority over seed.

src/tensorwaves/data/tf_phasespace.py
```diff
@@ -25,21 +28,31 @@ def __init__(
...
+        seed = None
+        if isinstance(seed_or_rng, UniformRealNumberGenerator):
+            seed = seed_or_rng.seed
+        elif isinstance(seed_or_rng, int):
+            seed = seed_or_rng
+        else:
+            raise NotImplementedError(
+                f"Cannot work with a seed or RNG of type {seed_or_rng.__class__.__name__}"
+            )
+        weights, particles = self.phsp_gen.generate(n_events=size, seed=seed)
```
spflueger, line 44: I'm a bit puzzled why the seed is given to the generate call. Since this is called several times due to the bunching, every bunch is created with the same seed

spflueger (changes requested):
I only looked at the important parts of the code for now. I think its best discussed in a short meeting

spflueger (approved):
LGTM

PR #173, opened 2020-11-20
feat: add CSVSummary callback

sebastianJaeger:
Added functionality to save and append intermediate results to a csv value to have a history of parameter values. Contributes to #159


src/tensorwaves/optimizer/callbacks.py
```diff
@@ -84,6 +85,48 @@ def finalize(self) -> None:
...
+
+    def __call__(self, parameters: dict, estimator_value: float) -> None:
+        self.__function_call += 1
+        if self.__function_call % self.__step_size != 0:
+            return
+        output_dict = {
+            "Time": datetime.now(),
+            "Iteration": self.__function_call,
+            "Estimator_Type": self.__estimator_type,
+            "Estimator_Value": float(estimator_value),
```
redeboer, line 112: Just a naming detail: I would either use `PascalCase` or `snake_case`, not a mix of both.

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -84,6 +85,48 @@ def finalize(self) -> None:
...
+            # name: float(value) for name, value in parameters.items()
```
redeboer, line 113: Comment can be removed?

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -6,6 +6,7 @@
 
 import tensorflow as tf
 import yaml
+from pandas import DataFrame
```
redeboer, line 9: I would use `import pandas as pd` as to distinguish namespaces in the rest of the code below.

redeboer (changes requested):

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -84,6 +85,48 @@ def finalize(self) -> None:
...
+            # name: float(value) for name, value in parameters.items()
+        }
+        for name, value in parameters.items():
+            output_dict[name] = float(value)
+
+        df = DataFrame(output_dict, index=[self.__function_call])
+        df.to_csv(
+            self.__stream,
+            mode="a",
+            header=self.__function_call == 10,
```
spflueger, line 122: This looks a bit dirty. It would be best to do the initialization in the init of the Callback.

redeboer: This is how it's done in the YAMLWriter
https://github.com/ComPWA/tensorwaves/blob/d5dedc77fc16ca41da3f31a47280e8e7fb5c178f/src%2Ftensorwaves%2Foptimizer%2Fcallbacks.py#L155-L157

redeboer: Oh sorry nvm

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -84,6 +85,48 @@ def finalize(self) -> None:
...
+            "Estimator_Value": float(estimator_value),
+            # name: float(value) for name, value in parameters.items()
+        }
+        for name, value in parameters.items():
+            output_dict[name] = float(value)
+
+        df = DataFrame(output_dict, index=[self.__function_call])
+        df.to_csv(
+            self.__stream,
+            mode="a",
```
redeboer, line 120: There's a bug here actually: if you use a file that already exists, this will keep appending to the old one. Therefore it's better to (like @spflueger pointed out) create one `DataFrame` in the `__init__`.

redeboer: So in the `__init__` I would do:
```python
data_frame.to_csv(
    self.__stream,
    header=True,
    index=False,
)
```
then in the `__call__`:

```python
data_frame.to_csv(
    self.__stream,
    header=False,
    index=False,
)
```

(Still not really sure though whether I like this writing through a `DataFrame`.)

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -84,6 +85,48 @@ def finalize(self) -> None:
...
+        if self.__function_call % self.__step_size != 0:
+            return
+        output_dict = {
+            "Time": datetime.now(),
+            "Iteration": self.__function_call,
+            "Estimator_Type": self.__estimator_type,
+            "Estimator_Value": float(estimator_value),
+            # name: float(value) for name, value in parameters.items()
+        }
+        for name, value in parameters.items():
```
redeboer, line 114:
```
output_dict.update({name: float(value) for name, value parameters.items()})
```

redeboer (approved):

PR #177, opened 2020-11-23
feat: implement Loadable callback

redeboer:


src/tensorwaves/optimizer/callbacks.py
```diff
@@ -2,121 +2,85 @@
...
 import tensorflow as tf
 import yaml
-from tqdm import tqdm
 
-from tensorwaves.interfaces import Estimator
+
+class Loadable(ABC):
+    @staticmethod
+    @abstractmethod
+    def load_latest_parameters(filename: str) -> dict:
```
spflueger, line 15: why not make it
```python3
@abstractmethod
￼def latest_parameters() -> dict:
```

src/tensorwaves/optimizer/callbacks.py
```diff
@@ -150,48 +122,63 @@ def __init__(
...
+        )
 
-    def finalize(self) -> None:
-        for callback in self.__callbacks:
-            callback.finalize()
+    def on_function_call_end(self) -> None:
+        self.__stream.close()
+
+    @staticmethod
+    def load_latest_parameters(filename: str) -> dict:
```
spflueger, line 178: With the different interface suggestion, the loading of the state from the resource would be handled by some static function like:

```python3

@staticmethod
def load_from_file(filename:str) -> "YamlSummary":
...
```
while the usage would be
```python3
summary = YamlSummary.load_from_file("asdf.yml")
params = summary.latest_parameters()
```

src/tensorwaves/optimizer/callbacks.py
redeboer: Because it actually loads the latest parameters. The stream that's used by `on_iteration_end` is write mode and is also closed after `on_function_call_end` is called. Note also that `CSVSummary` (and eventually `TFSummary`) have a different implementation that does not use Python `open` (internally a stream).

spflueger: why cant the loading be done in a separate function that is not part of the "interface"?

src/tensorwaves/optimizer/callbacks.py
redeboer: I tried to implement your suggestions, but indeed the mixing of read and write mode messes up the tests and I would have to add even more logics. Suggests bad design. (Not saying the current design is good either, but at least avoids this need for additional logic in the implementation.

src/tensorwaves/optimizer/callbacks.py
spflueger: Ok, so the summary is that the `Loadable` interface is not crucial, as its not connecting modules together, but "only" to the user (UI). Hence there are no avalanches propagating through the code if this will be changed later on. For more flexibility (e.g: writing to some web based API), the current interface would have to be abolished. The tricky part when having a interface like `def latest_parameters() -> dict:` is that both the input and output are connected to the state of the class. This can be tricky when loading a file in read mode, but writing to it later on.

spflueger (approved):

Issue #162, opened 2020-11-13
Isolate dynamics out from amplitude module


redeboer:
For example:
https://github.com/ComPWA/tensorwaves/blob/7c95713a509c3cb0a91e31103fb7d4be7e7ecdc3/src/tensorwaves/physics/helicity_formalism/amplitude.py#L604-L646

is not specific to the helicity formalism and should be placed under a new module like `tensorwaves.physics.dynamics`

spflueger:
Note this is handled with the design change to use sympy expression for the model description #203

PR #180, opened 2020-11-26
fix: remove default dynamics

redeboer:
Implements https://github.com/ComPWA/expertsystem/pull/395 (release 0.6.4) and removes the default dynamics.
https://github.com/ComPWA/tensorwaves/blob/ca494ea48a552b65cf1de40fb88e188ca20450dc/src%2Ftensorwaves%2Fphysics%2Fhelicity_formalism%2Famplitude.py#L159-L160

That was an implementation that seemed convenient, but modified the dynamics section of the input `AmplitudeModel`. 

Note that a long-term fix should come through #172 and https://github.com/ComPWA/expertsystem/issues/22


redeboer:
Note: 10e115e has to be reverted once https://github.com/ComPWA/expertsystem/pull/395 is through and `expertsystem==0.6.4` is released

tests/recipe/test_amplitude_creation.py
```diff
@@ -33,7 +33,7 @@ def test_helicity(helicity_model: es.AmplitudeModel):
 
     builder = IntensityBuilder(model.particles, kinematics, phsp_sample)
     intensity = builder.create_intensity(model)
-    assert len(intensity.parameters) == 9
+    assert len(intensity.parameters) == 10
```
spflueger, line 36: why the parameters go up by 1?

redeboer: Because the meson radius is written as a parameter as well now

redeboer: Oh no, forget that: d2d1b90
This also came by in #159, I think

redeboer: This makes it a bit more transparent c666dd8

tests/recipe/test_amplitude_creation.py
```diff
@@ -33,7 +33,17 @@ def test_helicity(helicity_model: es.AmplitudeModel):
 
     builder = IntensityBuilder(model.particles, kinematics, phsp_sample)
     intensity = builder.create_intensity(model)
-    assert len(intensity.parameters) == 9
+    assert set(intensity.parameters) == {
+        "strength_incoherent",
+        "Position_J/psi(1S)",
+        "Width_J/psi(1S)",
```
spflueger, line 39: Is it intended that the J/psi has a BreitWigner dynamics?

redeboer: I wouldn't say intended, but currently, mass and width are always registered, no matter what the dynamics are...
https://github.com/ComPWA/tensorwaves/blob/d2d1b90a26b532bc16a132f538a43153e2e87426/src%2Ftensorwaves%2Fphysics%2Fhelicity_formalism%2Famplitude.py#L591-L632

I would say this should be properly tackled in e.g. #162. This PR just removes the default dynamics (particularly the buggy way in which it was implemented).

spflueger (approved):

PR #182, opened 2020-12-09
test: add angular distribution tests from pycompwa

### LINKED ISSUE #181: Port angular distribution checks from pycompwa
Port the tests from pycompwa which verify the correctness of the angular parts of the helicity formalism
https://github.com/ComPWA/pycompwa/tree/master/tests/angular-distribution-tests

spflueger:
First test is ported from `pycompwa`, together with some fixture code.

I also made a small change to the variable names, as they were not consistent.

The second one I will tackle tomorrow. But this will serve as the baseline for all other distribution test.
@redeboer @Leongrim @sebastianJaeger , take a look if you like the test setup. Ideally I would like you to possibly extend this with your channels.

Closes #181

TODOs:

- [x] add second pycompwa distribution check
- [x] add scipy as a test dependency


tests/physics/helicity_formalism/test_amplitude.py
```diff
@@ -77,7 +78,11 @@ def test_coherent_intensity(functions, test_data, expected_results):
 def test_coefficient_amplitude(
     function, mag, phase, test_data, expected_results
 ):
-    model = _CoefficientAmplitude(function, mag, phase)
+    model = _CoefficientAmplitude(
+        function,
+        tf.constant(mag, dtype=tf.float64),
```
redeboer, line 83: Do we leave the responsibility of chooseing the `dtype` to the caller?

tests/physics/helicity_formalism/angular_distributions/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+from matplotlib import pyplot as plt
+
+from tensorwaves.data.generate import generate_data
+from tensorwaves.physics.helicity_formalism.amplitude import IntensityBuilder
+from tensorwaves.physics.helicity_formalism.kinematics import (
+    HelicityKinematics,
+)
+
+
+class Histogram:
```
redeboer, line 20: How about using `attr.s`?

spflueger:
I would just keep this PR as it is and @Leongrim  and @sebastianJaeger  can add their channels in follow up PRs. It also reflects the title this way.

tests/physics/helicity_formalism/angular_distributions/epem_to_dm_d0_pip_decay/test_distributions.py
```diff
@@ -0,0 +1,160 @@
...
+from typing import Any, Callable, List, Optional, Tuple
+
+import numpy as np
+import pytest
+
+
+# Use this function to reproduce the model file.
+# Note the normalization part has been removed!
+def generate_model() -> None:
+    from expertsystem.amplitude.helicity_decay import (
```
redeboer, line 15: Why inline imports?

tests/physics/helicity_formalism/angular_distributions/epem_to_dm_d0_pip_decay/test_distributions.py
```diff
@@ -0,0 +1,160 @@
...
+        final_state=[("D0", [0]), ("D-", [0]), ("pi+", [0])],
+        allowed_intermediate_particles=["D(2)*(2460)+"],
+        allowed_interaction_types="em",
+        particles=particles,
+    )
+
+    generator = HelicityAmplitudeGenerator()
+    amplitude_model = generator.generate(result)
+    amplitude_model.dynamics.set_non_dynamic("D(2)*(2460)+")
+    write(amplitude_model, "model.yml")
```
redeboer, line 44: Better use
```python
write(amplitude_model, output_dir + "model.yml")
```
and import that fixture with
```python
def generate_model(output_dir: str) -> None:
```

redeboer: but: why is this model also committed?

tests/physics/helicity_formalism/angular_distributions/epem_to_dm_d0_pip_decay/model.yml
```diff
@@ -0,0 +1,172 @@
+Kinematics:
```
redeboer, line 1: Isn't it better to generate this model? That would
1. allow for testing more channels with the same test procedure (and therefore also avoid having all these `__init__.py` files)
2. avoid having to rewrite the dumped model once the `expertsystem` changes its dump format
3. avoid writing to disk (which causes issues sometimes when running `pytest -n auto`)

tests/physics/helicity_formalism/angular_distributions/d1_to_d0_pi_pi_decay/__init__.py
```diff
@@ -0,0 +1 @@
+"""Required to set mypy options for the tests folder."""
```
redeboer, line 1: Why put these tests in different folders (see comments about committing the `model.yml` files).

redeboer (commented):
Good job! Just have some comments about the structure of the `angular_distributions` folder

tests/physics/helicity_formalism/angular_distributions/epem_to_dm_d0_pip_decay/test_distributions.py
spflueger: This function is some additional piece of info that is not required for the test. I just included it so that others understand how I created that model file that is shipped. Furthermore I wanted to make clear what is needed for running that function.... Maybe its better to just drop that function....

tests/physics/helicity_formalism/angular_distributions/epem_to_dm_d0_pip_decay/test_distributions.py
spflueger: See also the comment above.

I want the test to be as precise as possible, hence I keep the input fixed. So that if this test fails, it cannot be due to some change in the created model file. Instead the amplitude building or the data generation can only be responsible.

tests/physics/helicity_formalism/test_amplitude.py
spflueger: With the introduction of the prefactor, the `_CoefficientAmplitude` expect, the magnitude and phase to be `float64`. Maybe the datatype can also be inferred from the magnitude and phase and use the same type for the prefactor to achieve higher compatiblity. For the amplitudes generated by the builder, everything is `float64` though.

tests/physics/helicity_formalism/angular_distributions/d1_to_d0_pi_pi_decay/__init__.py
redeboer: ac910b9

tests/physics/helicity_formalism/angular_distributions/epem_to_dm_d0_pip_decay/test_distributions.py
redeboer: In the `expertsystem` we want to move away from reading/writing in tests, because that causes issues when running tests in parallel, see e.g. https://github.com/ComPWA/expertsystem/issues/385#issuecomment-734440965. In addition, I think what are testing here is rather the I/O of the `expertsystem` (internally, `tensorwaves` just uses the `AmplitudeModel` class and uses `expertsystem.io` to read/write).

tests/physics/helicity_formalism/angular_distributions/epem_to_dm_d0_pip_decay/test_distributions.py
redeboer: Yes I think it should be clear from from the notebooks how to generate a model.

tests/physics/helicity_formalism/angular_distributions/conftest.py
```diff
@@ -0,0 +1,255 @@
...
+def __chisquare(
+    values: Sequence[float], errors: Sequence[float], expected: Sequence[float]
+) -> float:
+    return np.sum(
+        [((x[0] - x[1]) / x[2]) ** 2 for x in zip(values, expected, errors)]
+    )
+
+
+@pytest.fixture(scope="module")
+def chisquare_test() -> Callable[[Histogram, Callable], None]:
```
redeboer, line 45: I wonder if it isn't more transparent and convenient to merge the `angular_distributions` folder into one `test_angular_distributions.py` test file that contains the function definitions that have to be defined as fixtures under `conftest.py`.

tests/physics/helicity_formalism/test_angular_distributions.py
```diff
@@ -0,0 +1,538 @@
...
+            "Magnitude_D(1)(2420)0_to_D*(2010)+_0+pi-_0;D*(2010)+_to_D0_0+pi+_0;"
+        ].value = 0.5
+
+        es.io.write(
+            amplitude_model, f"{SCRIPT_DIR}/{self.__class__.__name__}.yml"
+        )
+
+    # Use this function to reproduce the theoretical predictions.
+    @staticmethod
+    def calc_distributions() -> List[Tuple[str, Any]]:
```
redeboer, line 169: I'm not sure about these `calc_distributions` methods. What I did now for `generate_model` is to run those from the `__main__` for this script, so that it's easier to update the committed model files once the `expertsystem` changes. But these `calc_distributions` methods are not easily called because I'm not sure what to do with their output.

redeboer (approved):
So apart from , I think we can move on with this PR

tests/physics/helicity_formalism/test_angular_distributions.py
redeboer: 376011f

PR #187, opened 2021-01-05
fix!: use es.FormFactor in _RelativisticBreitWigner

redeboer:
Small fix for a bug that was forgotten to be changed in #134 (the change was correctly implemented for `_NonDynamic`
https://github.com/ComPWA/tensorwaves/blob/c35f270bc6a119bcb0bc2525eca1c24bb97c45ed/src/tensorwaves/physics/helicity_formalism/amplitude.py#L456-L461

Not sure why this wasn't spotted before. Long-term, this is better addressed through #162.

## To-do list
- [x] Correctly import form factor from `AmplitudeModel` for a relativistic BW
- [x] Update test values for the new model (deterministic test and callback test)
- [x] Revisit fit example of the example workflow


redeboer:
@spflueger It seems that the pre-factor for relativistic Breit-Wigners was converted incorrectly form the `AmplitudeModel` so far. This bug fix causes some of the tests to break again, but the model also looks completely different. Compare the models in the workflow notebooks prior to optimizing:

**stable branch** (see [here](https://pwa.readthedocs.io/projects/tensorwaves/en/0.1.3/usage/3_perform_fit.html#optimize-the-intensity-model))
image

**this branch** (generated locally with c35f270)
image

Is this the expected behaviour when adding a form-factor?

redeboer:
https://github.com/ComPWA/tensorwaves/pull/187#issuecomment-754752089 So actually there are three separate problems:
1. the check on whether the correct formalism is used was faulty (fixed by #188)
2. the form factor was imported incorrectly from the `es.AmplitudeModel` (this PR #187).
3. there is a factor missing in the relativistic BW without form factor:
   https://github.com/ComPWA/tensorwaves/blob/c35f270bc6a119bcb0bc2525eca1c24bb97c45ed/src/tensorwaves/physics/helicity_formalism/amplitude.py#L421-L425
   should have mass and width, like here
   https://github.com/ComPWA/tensorwaves/blob/c35f270bc6a119bcb0bc2525eca1c24bb97c45ed/src/tensorwaves/physics/helicity_formalism/amplitude.py#L451-L453

These three problems can be addressed by separate PRs (more transparent). This PR will need to fix some of the test values and rethink the example parameter values of the workflow notebook.

redeboer:
22d5eba fixes the tests, apart form `test_callbacks.py`. That test has an additional problem: it checks whether the parameters that the optimizer produces after the optimization is finished are equal to the 'latest' parameters stored to the traceback. However, (1) the callbacks by default store write the latest parameters every 10 steps and (2) the values are not written on the last step. So the "latest" parameters in the files that the callbacks (YAML and CSV) dump are actually not the latest parameters.

I can fix this tests quickly, but we may have to think of a fix here (@spflueger)?

PR #188, opened 2021-01-06
fix: throw invalid L error correctly

spflueger:
- [x] Throw exception correctly when using helicity formalism
- [x] Test if the exception is raised as expected


spflueger:
We could add a small test, that makes sure this error is thrown when this case actually appears.

spflueger:
Ok done. Unfortunately the design of the `IntensityBuilder` makes the testing quite difficult. Hence the rather abstruse test. Improving that design can be put on our Todo list.

redeboer (approved):
Nice test! Although the syntax can indeed be improved, hopefully through #162

PR #189, opened 2021-01-06
fix: add mass*width factor to relBW w/o ff

redeboer:
Adds a factor of _mass_ * _width_ to the relativistic Breit-Wigner **without** form factor as well. Previously, this only happened for the rel. BW with form factor.

Note: this more or less restores the distribution of the example workflow to the first distribution shown in https://github.com/ComPWA/tensorwaves/pull/187#issuecomment-754752089.

To summarize:
1. 574fe16 (v0.1.3, before with FF fix of #187)
   image

2. c35f270 (after #187)
   image

3. 1f7d1be (this PR)
   image


spflueger (approved):
Looks good!

PR #192, opened 2021-01-07
refactor!: normalize function in NLL

### LINKED ISSUE #191: Missing Intensity normalization prevents fits
Amplitudes generated with the expertsystem cannot be used in fits as a normalization of the intensity is missing. Hence the expertsystem is not compatible with the current version of tensorwaves.

spflueger:
The expertsystem was changed to not normalize the Intensity automatically anymore. To adapt to this change, the NLL estimator now gets another required argument, a phase space dataset. This is used to normalize the function.

Closes #191

### DIFF
--- a/src/tensorwaves/estimator.py
+++ b/src/tensorwaves/estimator.py
@@ -2,10 +2,42 @@
 
 All estimators have to implement the `~.interfaces.Estimator` interface.
 """
+from typing import Dict
+
 import numpy as np
 import tensorflow as tf
 
 from tensorwaves.interfaces import Estimator, Function
+
+
+class _NormalizedFunction(Function):
+    def __init__(
+        self,
+        unnormalized_function: Function,
+        norm_dataset: dict,
+        norm_volume: float = 1.0,
+    ) -> None:
+        self._model = unnormalized_function
+        # it is crucial to convert the input data to tensors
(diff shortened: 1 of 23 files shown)


src/tensorwaves/estimator.py
```diff
@@ -2,24 +2,63 @@
...
+        phsp_set: A phase space dataset, which is used for the normalization.
+            The model has to be evaluateable with this dataset. When correcting
+            for the detector efficiency use a phase space sample, that passed
+            the detector reconstruction.
 
     """
 
-    def __init__(self, model: Function, dataset: dict) -> None:
-        self.__model = model
+    def __init__(self, model: Function, dataset: dict, phsp_set: dict) -> None:
```
redeboer, line 57: If I understand correctly, this follows the interface of ComPWA, see here
https://compwa.github.io/ComPWA/classComPWA_1_1Estimator_1_1MinLogLH.html

What was the reason not to have `phsp_set` as a required argument for this estimator in `tensorwaves`?

spflueger: Good quesion, unfortunately I did not write down my reasoning back then. I guess simplicity, for the estimator. Naively speaking an estimator is supposed to compare a function to a dataset. This phase space sample is not exactly intuative.

redeboer (approved):
LGTM

PR #194, opened 2021-01-07
test: check shape of kinematics dataset

redeboer:
The doc CI for #192 fails because the shape of the data set that is returned by `HelicityKinematics.convert` is incorrect: its values contain floats, and should all be lists. This PR adds a test that checks the shape.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -42,12 +42,12 @@ def output_dir(pytestconfig) -> str:
 
 @pytest.fixture(scope="session")
 def helicity_model() -> AmplitudeModel:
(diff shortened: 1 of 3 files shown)


redeboer:
@spflueger The test will fail here too. Strange though that a dataset is correctly generated for the example notebook.

redeboer:
Note: I assume the shape of a data set (a `dict` of `ndarray`s) to be such that it can be directly converted into a `pandas.DataFrame`, like happens here:
https://pwa.readthedocs.io/projects/tensorwaves/en/0.1.3/usage/2_generate_data.html#visualize-kinematic-variables

Issue #184, opened 2020-12-18
Extract Intensity components from total intensity


spflueger:
This is required when calculating fit fractions and visualizing different components of the full intensity.

PR #222, opened 2021-03-10
refactor!: implement Model interface to lambdify Callables

### LINKED ISSUE #37: Verify helicity angle test
The last subsystem in the helicity angle test shows sign switch for phi (phi -> -phi)

Check if this is valid! The Wigner D function has to show this symmetry for it to be valid.

**[Estimated effort: 1-3 days]**

Note PR #203 will move kinematics to the expertsystem. The recursive angle calculation is well defined and allows for correct phi calculations. Hence the subsystem support is dropped in #203

### LINKED ISSUE #45: Move SubSystem class to physics or formalism module
Currently, the `SubSystem` class is only implemented in the `helicityformalism` class, but it will probably also be needed by other formalisms.

_Originally posted by @redeboer in https://github.com/ComPWA/tensorwaves/pull/39#discussion_r415800915_

### LINKED ISSUE #64: Implement gradient for UnbinnedNLL
Implement the analytic gradient.

- [ ] Clarify how to use the gradient in the Estimator with Intensity which hides the backend (I think this comes back to the "ComPWA math language"). So currently I think it is necessary to have a graph structure that describes the computation, which can be converted to an actual computation. Same concept that tensorflow follows. This would solve many problems at once.
- [ ] Actually implement the gradient

Some tensorflow specific code might look like this (copied from amplitf):
```python
for i, p in enumerate(float_pars):
     p.update(par[i])
with tf.GradientTape() as gradient:
    gradient.watch(float_pars)
    nll_val = nll(*args)
    g = gradient.gradient(
        nll_val, float_pars, unconnected_gradients=tf.UnconnectedGradients.ZERO)
    g_val = [i.numpy() for i in g]
    return g_val
```

### LINKED ISSUE #70: Improve helicity formalism amplitude builder
The the current design is that the "builder functions" (`_create_some_class`) are strongly coupled to the builder and are part of the recursion. In other words the `_create_some_class` function has to make calls to `create_element` in the builder for its children. That's not good.
So it would be better if the recursion is done completely inside the builder, and an element is constructed by a `_create_some_class` function by receiving its children directly.

### LINKED ISSUE #162: Isolate dynamics out from amplitude module

spflueger:
**Several breaking changes**

All physics (`amplitude`, `dynamics`, `kinematics`) has been moved to the `expertsystem`, where the amplitude model is now formulated in terms of SymPy. TensorWaves is now more abstract: the core has become a template `Model` class that can be 'lambdified' into a `Function` with different computational backends. The `esimator` and `data` modules are still there, but changed slightly because of the intermediate `Model` layer.

Since the responsibilities of `tensorwaves` have changed quite a bit and much code has been removed, the module structure has been changed drastically as well. The example workflow notebooks therefore changed as well.

See also: https://github.com/ComPWA/expertsystem/pull/454
Reopened from https://github.com/ComPWA/tensorwaves/pull/203

This PR also closes several issues, because the physics has been 'abstracted out' to the `expertsystem` through the `Model` interface.

~#184~ (will be illustrated in a notebook in a follow-up PR)
Closes #172
Closes #162
Closes #146
Closes #70
Closes #64
Closes #45
Closes #37


src/tensorwaves/data/generate.py
```diff
@@ -12,64 +14,58 @@
...
 
 
 def _generate_data_bunch(
     bunch_size: int,
     phsp_generator: PhaseSpaceGenerator,
     random_generator: UniformRealNumberGenerator,
     intensity: Function,
-    kinematics: Kinematics,
-) -> np.ndarray:
+    kinematics: HelicityKinematics,
```
spflueger, line 27: This should be a Kinematics interface. If we do not have an interface for that, then one has to be created

src/tensorwaves/data/generate.py
```diff
@@ -12,64 +14,58 @@
...
     kinematics: HelicityKinematics,
     intensity: Function,
     phsp_generator: Callable[
-        [ParticleReactionKinematicsInfo], PhaseSpaceGenerator
+        [ReactionInfo], PhaseSpaceGenerator
     ] = TFPhaseSpaceGenerator,
     random_generator: Optional[UniformRealNumberGenerator] = None,
     bunch_size: int = 50000,
-) -> np.ndarray:
+) -> MomentumPool:
```
spflueger, line 53: I'm not a huge fan of the choice `MomentumPool` here (I know this was kind of brought to live by myself) My reasoning is that Pool suggest some kind of design (https://en.wikipedia.org/wiki/Pool_(computer_science)) and here its simply a dictionary of momenta or something in that direction

setup.cfg
```diff
@@ -42,7 +42,7 @@ setup_requires =
     setuptools_scm
 install_requires =
     amplitf
-    expertsystem == 0.6.9
+    expertsystem @ git+https://github.com/ComPWA/expertsystem@sympy
```
spflueger, line 45: We have to remember to change this back and update before the merge

src/tensorwaves/physics/amplitude.py
```diff
@@ -0,0 +1,210 @@
...
+
+    return backend
+
+
+class LambdifiedFunction(Function):
+    def __init__(
+        self,
+        function: Callable,
+        argument_keys: Tuple[str, ...],
+        parameters: Optional[Mapping[str, Union[complex, float]]] = None,
```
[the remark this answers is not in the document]
redeboer: That would mean `Model.lambdify` has to return an arbitrary callable, not a `Function` (in this case: `SympyModel.lambdify` and `LambdifiedFunction` resp.). See here specifically
https://github.com/ComPWA/tensorwaves/blob/66330b09c22eaa8a949d5d785f35fb746d7cc3bd/src/tensorwaves/physics/amplitude.py#L192-L196

src/tensorwaves/interfaces.py
```diff
@@ -10,64 +12,91 @@ class Function(ABC):
...
-    separation is to facilitate the events when parameters have changed.
+    Hence while the domain variables are the argument of the evaluation (see
+    :func:`~Function.__call__`), the parameters are controlled via a getter and
+    setter (see :func:`~Function.parameters`). The reason for this separation
+    is to facilitate the events when parameters have changed.
     """
 
     @abstractmethod
-    def __call__(self, dataset: dict) -> list:
+    def __call__(self, dataset: DataSet) -> ScalarSequence:
```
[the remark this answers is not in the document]
redeboer: Switched to type aliases for now, then these can be improved upon later on
https://github.com/ComPWA/tensorwaves/blob/7e7d0da9ea592de297f0844bedb59e324f49d2a1/src/tensorwaves/interfaces.py#L20-L22

setup.cfg
redeboer: Requires a (pre-)release of https://github.com/ComPWA/expertsystem/pull/454 first

tests/data/test_generate.py
```diff
@@ -1,138 +1,160 @@
...
-            [-0.12901200642, 0.8756471174, -0.77276118186, 1.18270053384],
-            [-0.04750914038, 0.01974185344, -0.92886918322, 0.94003380539],
-            [-0.67129360197, 0.03043449972, 0.88545452402, 1.11973649108],
-            [0.68138748966, -1.08237744503, -0.3566346599, 1.33462985945],
+        1: [
+            [1.35290167494, 0.86789927359, 1.01492686264, 0.16970126666],
+            [0.88914805719, -0.19071913356, 0.08713876399, -0.85346269922],
+            [0.76423429742, 0.15644538516, -0.65782537005, 0.32958450934],
+            [0.69736757037, -0.04948850316, -0.10956533074, 0.67353479347],
+            [1.46945697438, -0.63259191506, -1.06583460163, 0.77775904623],
```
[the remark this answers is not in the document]
redeboer: These are the values that were there prior to #187, see [`8b69eb82`](https://github.com/ComPWA/tensorwaves/commit/8b69eb82e73239822c78899de91be5639d2e70f6#diff-8cd0bee7a2b440b80d70a2f62968707d4c3c17f328e818e64501adb8fe10404d). So I assume that #187 introduced a bug.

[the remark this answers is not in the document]
spflueger: Ah yes. Ok then everything makes sense again

setup.cfg
redeboer: Fixed by 454980a

src/tensorwaves/data/generate.py
redeboer: Fixed by 6b8a559 and follow-up commits as well as a rename from `MomentumPool` to `EventCollection` in https://github.com/ComPWA/expertsystem/pull/454

src/tensorwaves/data/generate.py
redeboer: Fixed by 6b8a559 and follow-up commits

src/tensorwaves/physics/amplitude.py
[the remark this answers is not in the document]
redeboer: Fixed by 6b8a559 and follow-up commits

redeboer (approved):

PR #227, opened 2021-03-15
refactor: rename converter to transformer

redeboer:
Some renamings:
- `SympyUnbinnedNLL` --> `UnbinnedNLL`
- `data.adapter` --> `data.transform`
- `DataConverter.convert` --> `DataTransformer.transform`
- `HelicityKinematicsConverter.convert` --> `HelicityTransformer.transform`
- `kinematics` argument of `generate_data` --> `data_transformer`


src/tensorwaves/data/transform.py
```diff
@@ -1,14 +1,14 @@
...
 
-from tensorwaves.interfaces import DataConverter, DataSample
+from tensorwaves.interfaces import DataSample, DataTransformer
 
 
-class HelicityKinematicsConverter(DataConverter):
+class HelicityTransformer(DataTransformer):
     """Convert four-momentum tuples to variables in the helicity formalism.
 
-    Implementation of a `.DataConverter` based on the
```
spflueger, line 11: Shouldn't we also make the change `convert` -> `transform` in the interface definition?

spflueger (commented):
Looks good. I one comment about the naming of the DataTransformer interface

src/tensorwaves/data/transform.py
redeboer: That's right. I was also wondering whether the naming in the `expertsystem` should be changed:
https://pwa.readthedocs.io/projects/expertsystem/en/0.7.0a0/api/expertsystem.amplitude.kinematics.html#expertsystem.amplitude.kinematics.HelicityAdapter.convert

spflueger: I would say so

redeboer: https://github.com/ComPWA/expertsystem/pull/524

spflueger: Ok good. Right now its a bit of a mixture anyways, since the `HelicityAdapater` in the `expertsystem` does not follow the same interface `EventCollection` vs `Dict[str, np.ndarray]`

spflueger (approved):
GJ

PR #231, opened 2021-03-16
feat: add scipy Optimizer adapter

spflueger:


mayou36:
Hey, that looks nice, but I think we should have a chat again in order to avoid duplication (as we're adding an improved interface in [zfit right now](https://github.com/zfit/zfit/pull/287)

spflueger:
> Hey, that looks nice, but I think we should have a chat again in order to avoid duplication (as we're adding an improved interface in [zfit right now](https://github.com/zfit/zfit/pull/287)

Hey Jonas, thanks for the comment. Sure, how about next week? We still have to prepare some things for friday.

tests/optimizer/test_scipy.py
```diff
@@ -0,0 +1,99 @@
+from typing import Callable, Optional
+
+import pytest
+
+from tensorwaves.interfaces import Estimator
+
+try:
```
redeboer, line 7: Is this try-catch block needed in the tests? I would say to run the tests, one needs to have the developer or `requirements-test.txt`  dependencies installed anyway, because the tests are not installed.

tests/optimizer/test_parameter_flattener.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from tensorwaves.optimizer.minuit import ParameterFlattener
+from tensorwaves.optimizer._parameter import ParameterFlattener
```
redeboer, line 5: Good that `ParameterFlattener` is hidden now. I guess `get_backend_modules` and `gradient_creator` can also be hidden, but let's do that in a separate PR.

src/tensorwaves/optimizer/scipy.py
```diff
@@ -0,0 +1,105 @@
...
+        use_analytic_gradient: bool = False,
+        **scipy_options: Dict[Any, Any],
+    ) -> None:
+        self.__callback: Callback = CallbackList([])
+        if callback is not None:
+            self.__callback = callback
+        self.__use_gradient = use_analytic_gradient
+        self.__method = method
+        self.__minimize_options = scipy_options
+        # self.__minimize_options.update({"disp": True})
```
redeboer, line 37: Can this be removed?

setup.cfg
```diff
@@ -56,6 +56,10 @@ packages = find:
 package_dir =
     =src
 
+[options.extras_require]
```
redeboer, line 59: `amplitf` (above) can be removed from the dependencies. Also wonder if other dependencies like `iminuit` can be moved to the `extras_require`. But we had this in ComPWA as well: do we include Minuit (`iminuit`) by default?

Going a bit further: `pandas` is only needed for the `CSVSummary` callback. Could also become optional. So where to draw the line?

redeboer (approved):
GJ! Some inline comments, but I approve.
There are similarities between the `minuit` and `scipy` submodules now, we can think about how to clean that up in a follow-up PR.

tests/optimizer/test_scipy.py
spflueger: The thing is the test discovery will fail if this try except block is not there

src/tensorwaves/optimizer/scipy.py
spflueger: Ah true i wanted to check if this is good default behaviour...

setup.cfg
redeboer: Note: the pinned dependencies are currently unaffected by `extras_require`, but this will be addressed by @spflueger in a separate PR. https://github.com/ComPWA/tensorwaves/pull/231#discussion_r595892440 can also be addressed there.

spflueger: I think we should define some default packages tensorwaves works with and add those as requirements, so the user can do a complete workflow. Once the user wants extra features some additional packages have to be installed.

tests/optimizer/test_scipy.py
redeboer: But the only circumstance under which you want to discover tests is if you have `pytest` installed. And I think that should be done through installing `reqs/3.x/requirements-test.txt`.

setup.cfg
spflueger: So that means concretely minuit or scipy is a default (we can just stick with minuit i think). pandas would also be an install requirement if we choose it as a default, is csv summary default? I assume pandas will also be used for some extensions to the data module that might come later on

src/tensorwaves/optimizer/scipy.py
spflueger: db932dd659f4f383cf464643b0b36c6482fd0563

setup.cfg
spflueger: db932dd659f4f383cf464643b0b36c6482fd0563

tests/optimizer/test_scipy.py
spflueger: Thats true. Let me fix this then

spflueger: d41e7fc44dd46024c5684f859eac820578438848

PR #234, opened 2021-03-17
refactor: extend Callback interface

### LINKED ISSUE #230: Add an init and cleanup stage in the callback interface
Currently there is a unpleasent behavior in the optimization workflow using callbacks. When setting up callbacks for an optimization, the Callback is initialized during its construction (for example, opening up a file handle/stream). However when rerunning a fit, the file handle of the current callback instance would already be closed and a file IO error is thrown. As a workaround, the Callback has to be constructed again an handed to the new optimizer instance.

**Idea:**
To improve this, the Callback interface can be extended with a `init` and `cleanup` method that should be triggered before an optimization starts and after it terminated. Using these two new interfaces the correct behaviors can be implemented. (Note the names of the two methods are just an example).

spflueger:
Closes #230


redeboer (approved):
LGTM
As discussed, it would be better to tackle the jumps in the SciPy optimize process in this PR as well, but I currently don't see a solution.
image

Issue #14, opened 2020-04-17
Implement/port fit fraction functionality


spflueger:
It may be best to put this functionality under an overarching repository, see https://github.com/ComPWA/.github/issues/6.

For inspiration, see:
- https://compwa.github.io/usage/workflow/4_analyze_results.html#4.3-Calculate-fit-fractions
- [`pycompwa` code](https://github.com/ComPWA/pycompwa/blob/ee8b1d7fdba1643c5ee51eac8903650bcdf8a7c9/PyComPWA.cpp#L577-L589)
- [ComPWA code](https://github.com/ComPWA/ComPWA/blob/efabde557d63ebe5228dc2dd8e8f3986520c227a/Tools/FitFractions.cpp#L23-L85)

redeboer:
Done through wrapper repos like [ComPWA/jpsi-nstar](https://github.com/ComPWA/jpsi-nstar)

PR #244, opened 2021-03-22
feat: extract intensity components

### LINKED ISSUE #184: Extract Intensity components from total intensity

redeboer:
Closes #184


redeboer:
@spflueger Currently, the functions are only implemented in usage/step3 notebook. We have to think where and how this can be integrated into the package.

src/tensorwaves/model.py
```diff
@@ -109,13 +119,6 @@ def __init__(
         if not all(map(lambda p: isinstance(p, sp.Symbol), parameters)):
             raise TypeError(f"Not all parameters are of type {sp.Symbol}")
 
-        if not set(parameters) <= set(expression.free_symbols):
```
spflueger, line 112: why is this removed?

src/tensorwaves/model.py
```diff
@@ -180,3 +183,49 @@ def variables(self) -> FrozenSet[str]:
     @property
     def argument_order(self) -> Tuple[str, ...]:
         return tuple(x.name for x in self.__argument_order)
+
+
+def add_components(
```
spflueger, line 188: How about we put this into some new module, since this is quite special code related to the `HelicityModel`.

src/tensorwaves/model.py
redeboer: Because if you take a sub-expression that does not feature all the parameters that were in the original model, this will crash. I don't think a check like this is necessary here.

src/tensorwaves/model.py
redeboer: Yes, I suppose we need to re-introduce the `physics` module (but now separate from the `model` module). #15, #14, and #38 could be put there as well.

src/tensorwaves/model.py
redeboer: Possible, but then the caller always have to filter those beforehand useing `free_symbols`.

src/tensorwaves/model.py
spflueger: #15 and #38 belong together (model domain related)

#14 not related to these, but related to this PR. So we could def put those together. Maybe into `fit_fractions.py` or `components.py`.

I simply do not want `model.py` to contain model specific implementations.

src/tensorwaves/model.py
redeboer: Added the chack back but as warning e08f084

src/tensorwaves/model.py
redeboer: Ok, moved them to a separate (possibly temporary) `physics` module (b2a190d). The location of these functions will become clearer in through those other issues

Issue #100, opened 2020-06-26
Caching of partial calculations


spflueger:
The current Tensorflow code runs relatively slow (about 3x slower than pycompwa), which is related to the fact that tensorwaves is not caching any computations. Even the ones that that stay constant throughout the whole fitting/optimization phase (WignerD functions).

Some flexible and easy to use scheme is necessary to implement local caching.

**Update:** A lot of things have changed since. Now with the sympy + lambdify, the whole model can be optimized much easier. A separate method was added to the Model interface `performance_optimize`.

PR #245, opened 2021-03-23
feat: support lambdify to tensorflow

redeboer:
Many thanks to @mayou36 for pointing out [`tf.experimental.numpy`](https://www.tensorflow.org/api_docs/python/tf/experimental/numpy)!

Note: TF is currently the slowest of the backends for both optimizing and data generation. Could be that #100 offers a way out here and/or that we have to make the lambdifying a bit smarter.

### DIFF
--- a/src/tensorwaves/optimizer/minuit.py
+++ b/src/tensorwaves/optimizer/minuit.py
@@ -73,7 +73,7 @@ class Minuit2(Optimizer):
             update_parameters(pars)
             parameters = parameter_handler.unflatten(flattened_parameters)
             estimator_value = estimator(parameters)
-            progress_bar.set_postfix({"estimator": estimator_value})
+            progress_bar.set_postfix({"estimator": float(estimator_value)})
             progress_bar.update()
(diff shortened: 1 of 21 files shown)


src/tensorwaves/optimizer/minuit.py
```diff
@@ -73,7 +73,7 @@ def wrapped_function(pars: list) -> float:
             update_parameters(pars)
             parameters = parameter_handler.unflatten(flattened_parameters)
             estimator_value = estimator(parameters)
-            progress_bar.set_postfix({"estimator": estimator_value})
+            progress_bar.set_postfix({"estimator": float(estimator_value)})
```
spflueger, line 76: Why is this cast needed? Is the output from tf not pretty?

spflueger (approved):

src/tensorwaves/optimizer/minuit.py
redeboer: Yes, the estimator value is a `Tensor`, that looked weird in the progress bar

PR #247, opened 2021-03-24
feat: add caching of constant sub expressions in SympyModels

### LINKED ISSUE #100: Caching of partial calculations

spflueger:
Closes #100

### DIFF
--- a/src/tensorwaves/model.py
+++ b/src/tensorwaves/model.py
@@ -4,7 +4,7 @@ The `.model` module takes care of lambdifying mathematical expressions to
 computational backends. Currently, mathematical expressions are implemented
 as `sympy` expressions only.
 """
-
+# cspell: ignore xreplace
 import copy
 import logging
 from typing import Any, Callable, Dict, FrozenSet, Mapping, Tuple, Union
--- a/src/tensorwaves/model.py
+++ b/src/tensorwaves/model.py
@@ -146,7 +146,7 @@ class LambdifiedFunction(Function):
 class _ConstantSubExpressionSympyModel(Model):
     """Implements a performance optimized sympy based model.
 
-    Based on which symbols of the sympy expression are declared
+    Based on which symbols of the sympy expression are declared.
     """
 
     # pylint: disable=too-many-instance-attributes
--- a/src/tensorwaves/model.py
+++ b/src/tensorwaves/model.py
@@ -5,8 +5,9 @@ computational backends. Currently, mathematical expressions are implemented
 as `sympy` expressions only.
 """
 
+import copy
 import logging
-from typing import Callable, Dict, FrozenSet, Mapping, Tuple, Union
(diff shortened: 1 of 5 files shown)


spflueger:
I'm keeping it like this for now. Not the best user interface, but we have more issues regarding the Model interface and the gradient as well.

src/tensorwaves/model.py
```diff
@@ -77,6 +143,135 @@ def update_parameters(
...
+                input_args[non_fixed_arg_positions[i]] = x
+
+        def wrapper(*args: Tuple[Any, ...]) -> Any:
+            update_args(*args)
+            return lambdified_model(*input_args)
+
+        return wrapper
+
+    def performance_optimize(self, fix_inputs: DataSample) -> "Model":
+        return NotImplemented
```
redeboer, line 254: Shouldn't this method return the same as `SympyModel.performance_optimize()`? So returning another `_ConstantSubExpressionSympyModel`?

src/tensorwaves/model.py
```diff
@@ -77,6 +143,135 @@
...
+        if not is_constant:
+            for sub_expr in temp_constant_sub_expression:
+                placeholder = sp.Symbol(f"cached[{str(sub_expr)}]")
+                self.__constant_sub_expressions[placeholder] = sub_expr
+        return is_constant
+
+    def __replace_constant_sub_expressions(
+        self, expression: sp.Expr
+    ) -> sp.Expr:
+        new_expression = copy.deepcopy(expression)
```
redeboer, line 207: I'm surprised this `deepcopy` is necessary. Does `xreplace`/`subs` modify the expression?

redeboer (commented):
Good job!
I left a few comments. It's a bit obscure why the macos job keeps failing now, looking into it. This is a problem that sporadically happened before, so this has to be tackled anyway. (Still, it's strange that this PR makes the macos job more unstable.)

src/tensorwaves/model.py
spflueger: Good question. I'm actually not sure on the internals of the sympy expression, but the original expression not the be modified

src/tensorwaves/model.py
spflueger: I want this returned instance to implement the Model Interface

redeboer (approved):

redeboer:
A series of PRs related to the callbacks is coming up that might fix the CI problem seen here.

PR #251, opened 2021-03-26
fix: avoid cell magic in colab cell

redeboer:
Follow-up to #250. The cell magic conditionals didn't work well, which slowed down CI.


redeboer:
There's still another bug related to the custom hook (which create this colab cell): they do not work in the `tox -e sty` job. So eventually I want to move these hooks out again from the repo and put them under a separate ComPWA/pre-commit-hooks repository (which allows you to install the hook as a `repo` hook with `python` as language).
https://pre-commit.com/#new-hooks

PR #254, opened 2021-03-26
refactor: wrap fit result in class

### LINKED ISSUE #165: Define FitResult class
Some thoughts:
- The `dict` that is currently returned is not convenient from a user perspective and hard to debug
- See https://github.com/ComPWA/tensorwaves/pull/164#discussion_r524401478: it may solve the `parameters`, `estimator_value` problem
- Challenge will be to be general enough for different optimisers.

redeboer:
Wrapped the optimize `dict` output in a `FitResult` class. It's a rather simple `attr` data container that wraps what was previously in a `dict` form. Optimizer-specific info has been put under a `specifics` attribute.

### DIFF
--- a/src/tensorwaves/interfaces.py
+++ b/src/tensorwaves/interfaces.py
@@ -17,6 +17,11 @@ import attr
 import numpy as np
 from expertsystem.amplitude.kinematics import ReactionInfo
 
+try:
+    from IPython.lib.pretty import PrettyPrinter  # type: ignore
+except ImportError:
+    PrettyPrinter = Any
+
 # Data classes from the expertsystem do not work with jax and jit
 # https://github.com/google/jax/issues/3092
 # https://github.com/google/jax/issues/4416
@@ -135,6 +140,24 @@ class FitResult:  # pylint: disable=too-many-instance-attributes
     iterations: Optional[int] = None
     specifics: Optional[Any] = None
     """Any additional info provided by the specific optimizer."""
+
+    def _repr_pretty_(self, p: PrettyPrinter, cycle: bool) -> None:
+        class_name = type(self).__name__
+        if cycle:
+            p.text(f"{class_name}(...)")
+        else:
+            with p.group(indent=1, open=f"{class_name}("):
+                for field in attr.fields(type(self)):
+                    if field.name in {"specifics"}:
+                        continue
+                    value = getattr(self, field.name)
+                    if value != field.default:
+                        p.breakable()
+                        p.text(f"{field.name}=")
+                        p.pretty(value)
+                        p.text(",")
+            p.breakable()
+            p.text(")")
 
 
 class Optimizer(ABC):
--- a/src/tensorwaves/interfaces.py
+++ b/src/tensorwaves/interfaces.py
@@ -2,6 +2,7 @@
 
 from abc import ABC, abstractmethod
 from typing import (
+    Any,
     Callable,
     Dict,
     FrozenSet,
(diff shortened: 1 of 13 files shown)


src/tensorwaves/interfaces.py
```diff
@@ -111,27 +119,56 @@ class Estimator(ABC):
...
     def gradient(
-        self, parameters: Mapping[str, Union[float, complex]]
-    ) -> Dict[str, Union[float, complex]]:
+        self, parameters: Mapping[str, ParameterValue]
+    ) -> Dict[str, ParameterValue]:
         """Calculate gradient for given parameter mapping."""
 
 
+@attr.s(frozen=True, auto_attribs=True)
+class FitResult:  # pylint: disable=too-many-instance-attributes
```
redeboer, line 133: I hesitated to put the `FitResult` under `interfaces`, because it's not an abstract class. Alternative would be to put it under some new sub-module like `tensorwaves.optimizer.result` (sub-module to avoid cyclic imports in `minuit` and `scipy`.

redeboer:
One thing that's still missing is the covariance matrix. We have to think about a data type for that though.

src/tensorwaves/interfaces.py
spflueger: I think its fine there, since its part of the Optimizer interface

src/tensorwaves/optimizer/minuit.py
```diff
@@ -119,11 +119,12 @@ def wrapped_gradient(pars: list) -> Iterable[float]:
...
-            "execution_time": end_time - start_time,
-        }
+        return FitResult(
+            minimum_valid=minuit.valid,
+            execution_time=end_time - start_time,
+            function_calls=minuit.fmin.nfcn,
+            estimator_value=minuit.fmin.fval,
+            parameter_values=parameter_values,
+            parameter_errors=parameter_errors,
+            specifics=minuit,
```
spflueger, line 129: I don't like that the whole minuit object is attached to the fit result, because:

1. Its an object that does not represent only a fit result, but the whole minimizer
2. It contains duplicate information which is already available in the other attributes.

(see also comment on FitResult)

src/tensorwaves/interfaces.py
```diff
@@ -111,27 +119,56 @@ class Estimator(ABC):
...
+    minimum_valid: bool
+    execution_time: float
+    function_calls: int
+    estimator_value: float
+    parameter_values: Dict[str, ParameterValue]
+    parameter_errors: Optional[Dict[str, ParameterValue]] = None
+    iterations: Optional[int] = None
+    specifics: Optional[Any] = None
```
spflueger, line 141: I would either do one of these (both has its pros and cons):

1. not creating any class, and just return the optimizer/minimizer specific objects directly (for minuit a custom object would have to be created as it does not define one)
2. make the specifics into a dictionary of optional info

src/tensorwaves/optimizer/minuit.py
redeboer: Yes, that's true. As for 1., that's just the unfortunate design `iminuit` (the optimizer serves as a fit result container as well).

src/tensorwaves/interfaces.py
redeboer: Yes this is true. I would not favour 1, because it leaves the interface undefined (even though, internally, the package currently has no need for a `FitResult` interface).

redeboer:
So the way I see this PR now is that it _only_ (1) replaces the `dict` fit result with a class and (2) offers missing additional info (particularly the covariance matrix or hessian). This means that #165 is not closed.

Btw, the fact that the `dict` has been replaced with a class, helped fix one typo:
https://github.com/ComPWA/tensorwaves/blob/b2bb1f07abf8046f615ea4ededa7384b3ecba43b/src/tensorwaves/optimizer/scipy.py#L145

spflueger (approved):

PR #270, opened 2021-05-10
build: upgrade to AmpForm v0.9.x

redeboer:
Adds support for analytic continuation in particular.


redeboer:
Note that CI will fail due to https://github.com/ComPWA/ampform/pull/58. So requires AmpForm v0.9.1.

PR #311, opened 2021-08-30
fix: use correct version boundary syntax for pip

redeboer:
Fix-up to #310 

https://github.com/ComPWA/tensorwaves/blob/57f6ec0f04e0afc0cebcc955783d20cde2646be6/setup.cfg#L44

resulted in version conflicts downstream. AmpForm is now also correctly updated: 626abe1.


redeboer:
Analytic continuation notebook had to be removed (2cef8de). This will be reverted and fixed in #312.

PR #345, opened 2021-11-15
refactor: adapt implementation to AmpForm v0.12.x

redeboer:
Adapts the `data` module implementation so that it can work with [AmpForm v0.12.x](https://github.com/ComPWA/ampform/releases/tag/0.12.0). Closes https://github.com/ComPWA/ampform/issues/182.

Some major changes:
- Keys in a `DataSample` are now `str` only, not `int`. This means that momentum samples have keys `"p0"`, `"p1"`, etc. instead of `0`, `1`. This is in accordance with the `Symbol` names used in AmpForm's `HelicityModel.kinematic_variables`.
- Added a function `create_function()` that creates a `PositionalArgumentFunction`. This is a `Function` without parameters (compare `ParametrizedFunction`). The function and related class is used in the `DataTransformer` implementation, which doesn't need parameters.
- Data conversion can now be done with different back-ends.


redeboer:
95fa4c5 (#374) indeed seems to fix https://github.com/ComPWA/tensorwaves/pull/345#commitcomment-61398587 :tada:

Issue #323, opened 2021-09-21
Make it easier to inspect source-code of a BackendFunction


redeboer:
The lambdified function that is contained in `LambdifiedFunction` is 'hidden' by a dunder:

https://github.com/ComPWA/tensorwaves/blob/942f4b5b04f7e2c1d58b0ef09ba152e136e4db48/src/tensorwaves/model.py#L278

This makes it almost impossible to get the lambdified function or inspect the source code of that lambdified function. Note that simply making this attribute public does not entirely solve the problem: the lambdified function can be constructed by `optimized_lambdified`, meaning that it is still not possible to inspect the lambdified function even if you have access to this attribute.

See also #322

jonas-eschle:
I would suggest to make this (and the others) simply normal private attributes. Name mangling is usually just needed if there is a potential for naming conflicts, which I think is not given here?

Or, out of curiosity, why was this even been put to leading dunder?

spflueger:
I would ask the other way around: what's wrong with using a dunder by default? I think its quite similar to c++ private and protected. There you have the same ideal, to use private unless you really need protected or public. So my logic would be the same here, unless you need access to the variable make it a dunder.
So far I don't have any argument why using dunders by default is a bad thing.

jonas-eschle:
> I would ask the other way around: what's wrong with using a dunder by default? I think its quite similar to c++ private and protected.

Besides the Python convention, from a software design perspective: why private and not protected? If I would like to inherit from the class, should I not be able to access exactly these attributes? Like should not every `LambdifiedFunction`(in the OOP "is-a" sense) have this lambdified function accessible?

AFAIU, a single leading underscore is considered "private/protected", while a dunder is ["just" to avoid collisions](https://www.hacksoft.io/blog/underscores-dunders-and-everything-nice#name-mangling). So if you wanna do some real heavy magic or if the attribute is likely to collide with another attribute, that's a good use-case for dunder.

> There you have the same ideal, to use private unless you really need protected or public.

Python goes here more the path of "we're all adults here". There is no need to be over-protective with things, but rather about "communicating an intention". If anyone touches a private (`_`) attribute, the convention is clear: that is not guaranteed to be stable and may break anytime. We all know that. Making something `__` just makes sure that the attribute is not accessed by accident. Because *all* attributes can be accessed anyways. So using a dunder here conveis I think the wrong message:

[...]

PR #353, opened 2021-11-25
refactor!: create ParametrizedFunction interface

redeboer:
The existing `Function` interface has too many responsibilities: it (1) converts a `DataSample` to an array or tensor and (2) distinguishes parameter variables from domain variables. This worked fine so far, because `Function` was only used when optimizing, where you need to tweak those parameters. In #345 though, we need to a `Function` that only does (1), for the conversion of one `DataSample` (four-momenta) to another `DataSample` (kinematic variables) with different back-ends.

This PR creates a new `ParametrizedFunction` interface that does (1) and (2) and a 'bare' `Function` interface that does only (1). The input and output types of the `Function` interface are generic, so that it can be used in different types of function-like interfaces.

Related changes:
- Renamed `LambdifiedFunction` to `ParametrizedBackendFunction`.
- `DataTransformer.transform()` has become a `__call__` and `DataTransformer` derives from `Function`.
- `Estimator` also derives from `Function`: it computes an estimator value for a set of parameters.


redeboer:
@spflueger as discussed, `ParametrizedFunction` inherits from `Function`:

https://github.com/ComPWA/tensorwaves/blob/49eae48965d0c11e7a058580e77951ce9cd8b3a6/src/tensorwaves/interface.py#L32

This is probably fine with interfaces. Another option would be to give `ParametrizedFunction` a `function` attribute. This makes sense if `Function` were to have more attributes (for instance something like `expected_keys`, for the input `DataSample` that its call takes). In that case, composition in the interface of `ParametrizedFunction` may be safer.

src/tensorwaves/model/function.py
```diff
@@ -0,0 +1,117 @@
...
+    """Wrapper around a function with positional arguments.
+
+    This class provides a :meth:`__call__` that can take a `.DataSample` for a
+    function with `positional arguments
+    <https://docs.python.org/3/glossary.html#term-positional-argument>`_. Its
+    :attr:`argument_order` redirect the keys in the `.DataSample` to the
+    argument positions in its underlying :attr:`function`.
+    """
+
+    function: Callable[..., np.ndarray] = attr.ib(
```
spflueger, line 73: I dont quite understand why these have to be exposed?

src/tensorwaves/model/sympy.py
```diff
@@ -117,57 +119,45 @@ def optimized_lambdify(
...
-            sub_expression, symbols, backend, **kwargs
-        )
+        sub_function = lambdify(sub_expression, symbols, backend, **kwargs)
         sub_functions.append(sub_function)
 
     def recombined_function(*args: Any) -> Any:
-        new_args = [sub_function(*args) for sub_function in sub_functions]
-        return top_function(*new_args)
+        new_args = [func.function(*args) for func in sub_functions]
+        return top_function.function(*new_args)
```
spflueger, line 141: Ok I guess this answers my question above, but feels a bit like a design smell. Can't really follow the implementation details atm (not so up to date with the code details). So afaiu its about substituting parts of the function with sub functions and this takes care of the correct argument handling. Maybe you explain your thoughts

src/tensorwaves/model/sympy.py
```diff
@@ -414,16 +406,22 @@ def __init__(
         )
         self.max_complexity = max_complexity
 
-    def lambdify(self, backend: Union[str, tuple, dict]) -> Callable:
+    def lambdify(self, backend: Union[str, tuple, dict]) -> Function:
         """Lambdify the model using `~sympy.utilities.lambdify.lambdify`."""
-        return _sympy_lambdify(
+        if self.max_complexity is None:
```
spflueger, line 411: Why is this public btw? I could not find any doc string about this or quick reasoning to make sense of it

spflueger (commented):
Looks good & good job. Just two questions about some design choices

src/tensorwaves/model/function.py
redeboer: I try to use [`attrs`](https://www.attrs.org) where possible, because it encourages writing composite classes that are easy to understand. But since `attrs` has a [specific take on 'private' attributes](https://www.attrs.org/en/stable/init.html#private-attributes), there isn't really any way to hide the attribute. And I tend to go along with that when all attributes can be 'frozen' (a `dict` attribute, like in `ParametrizedBackendFunction` is a counter-example).

src/tensorwaves/model/sympy.py
redeboer: https://github.com/ComPWA/tensorwaves/pull/353#discussion_r757888265
The next question though is whether those public frozen attributes should be part of the interface. I decided not to, because the sole function of `Function` is to compute an array from a `DataSample`.

But indeed, the implementation here is a bit wonky. It seems related to #323: ideal would be to offer a mechanism to lambdify an expression tree to a 'function tree' in terms of that back-end, where it's possible to inspect sub-functions. But the `Function` interface does not suffice there, because you need some attributes that tell which sub-functions a top function provides.

src/tensorwaves/model/sympy.py
redeboer: Do you mean the `lambdify` method?

redeboer: Actually, while working on this PR, I started to have some doubts about the implementation classes. There is a lot of logics going on in the constructor. Would it be better to create some functions that build such classes? In this case, a `lambdify` **function** (could be a static method) that builds this `ParametrizedBackendFunction` from a SymPy expression? Currently, there is only one way customise the construction process, through the constructor.

spflueger: Ah sorry, I was talking about the `max_complexity`

redeboer:
Short summary of offline discussion:
- The `Model` interface can be removed (#357)
- It would be better to create a generic `Function` interface. The PR has been changed into that, so see the diff fore more details.
- Previous implementation details (see review comments) are only relevant in #345. They have been removed from this PR.

src/tensorwaves/model/sympy.py
redeboer: Fixed by #357

src/tensorwaves/model/sympy.py
redeboer: https://github.com/ComPWA/tensorwaves/pull/353#issuecomment-982770331

src/tensorwaves/model/function.py
redeboer: https://github.com/ComPWA/tensorwaves/pull/353#issuecomment-982770331

PR #359, opened 2021-12-01
ci: switch to nbmake

redeboer:
Switch from [`pytest-notebook`](https://pypi.org/project/pytest-notebook) to [`nbmake`](https://pypi.org/project/nbmake). Now it's again possible to run specific notebooks from the terminal with e.g.:

```shell
pytest --nbmake docs/usage/basics.ipynb
```

Other small fixes:
- Avoid `fast_lambdify()` in Jupyter notebooks to speed up docs workflow.
- Cast to `tuple` in `ParametrizedBackendFunction`.

### DIFF
--- a/docs/usage/step3.ipynb
+++ b/docs/usage/step3.ipynb
@@ -89,7 +89,6 @@
     "    expression=model.expression.doit(),\n",
     "    parameters=model.parameter_defaults,\n",
     "    backend=\"jax\",\n",
-    "    max_complexity=100,\n",
     ")"
    ]
   },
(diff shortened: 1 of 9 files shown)


docs/usage/step3.ipynb
```diff
@@ -89,7 +89,6 @@
     "    expression=model.expression.doit(),\n",
     "    parameters=model.parameter_defaults,\n",
     "    backend=\"jax\",\n",
-    "    max_complexity=100,\n",
```
redeboer, line 92: @Leongrim this may be of importance to your benchmark comparison. (See also #103)

PR #366, opened 2021-12-02
ci: avoid callbacks if unspecified

redeboer:
Previously, if no callback was specified in the optimizer constructor, an empty CallbackList would be created and on_optimize_end etc were always called. This is (theoretically) slower.

Some other improvements:
- Use [`attrs` next-generation API](https://www.attrs.org/en/stable/api.html#next-gen) (see also https://github.com/ComPWA/compwa-org/issues/90).
- Avoid creating stream on creation of Loadable callback.
- Fail pytest on warnings (this helped fishing out the above bug)

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -12,6 +12,7 @@ addopts =
     --ignore=docs/conf.py
     --ignore-glob=*/.ipynb_checkpoints/*
 filterwarnings =
+    error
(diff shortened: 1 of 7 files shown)


redeboer:
Slow-down mentioned in #367 seems to be caused by b0641fb. Fixed with ca3d6b2.

PR #399, opened 2022-01-11
feat: add minuit_modifier constructor argument

redeboer:
This makes it possible to access [`iminuit.Minuit`](https://iminuit.readthedocs.io/en/stable/reference.html#iminuit.Minuit)'s methods and attributes before calling `optimize()`.

### DIFF
--- a/docs/amplitude-analysis.ipynb
+++ b/docs/amplitude-analysis.ipynb
@@ -1310,7 +1310,7 @@
    "source": [
     ":::{seealso}\n",
     "\n",
-    "Call {meth}`~.Minuit2.setup` first in order to tweak the internal {attr}`~.Minuit2.iminuit` optimizer before calling {meth}`~.Minuit2.optimize`. See {ref}`usage/basics:Minuit2` for how to do this.\n",
+    "See {ref}`usage/basics:Minuit2` for how to tweak the internal {class}`iminuit.Minuit` optimizer.\n",
     "\n",
     ":::\n",
     "\n",
--- a/docs/amplitude-analysis.ipynb
+++ b/docs/amplitude-analysis.ipynb
@@ -1308,6 +1308,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    ":::{seealso}\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


redeboer:
@Leongrim, have a look if this helps.
@spflueger, I don't like the way `setup()` modifies the behaviour of `optimize()`, but this is the best I could come up with for now given this `Optimizer` interfaces.

spflueger:
Why does the minuit instance have to be tweaked before optimize? Cant that be done at construction?

A more general question. As far as I remember the operator interface is solely connected to the user. So a user can swap optimizers and does not have to adjust the optimize call. But the optimizer interface is not used in tensorwaves internally right?
Hence the interface is not really needed per se. Just some consistency thing for the user. But if this is too restrictive. Then I think it can be abandoned.

redeboer:
> Why does the minuit instance have to be tweaked before optimize? Cant that be done at construction?

I would have preferred that too, but one needs an estimator and the free parameters in order to construct an `iminuit.Minuit` instance. That would mean these become tied to the constructor, so each TensorWaves `Minuit2` instance is specific to each set of free parameters. (At the same time though, this `setup()` method just obscures that fact and one can mess up things by for instance inserting a different estimator in `setup()` than in `optimize()`...)

> A more general question. As far as I remember the operator interface is solely connected to the user. So a user can swap optimizers and does not have to adjust the optimize call. But the optimizer interface is not used in tensorwaves internally right? Hence the interface is not really needed per se. Just some consistency thing for the user. But if this is too restrictive. Then I think it can be abandoned.

[optimizer interface?] Yes, the `Optimizer` interface class is only used as an ABC to optimizer implementations and optimizer instances are not (currently) used elsewhere in the source code.

spflueger:
@redeboer Ok maybe we can do a quick call to discuss this.

redeboer:
> @redeboer Ok maybe we can do a quick call to discuss this.

@spflueger Done :heavy_check_mark: Usage can be previewed [here](https://tensorwaves--399.org.readthedocs.build/en/399/usage/basics.html#minuit2). Not the most pretty, but less error prone than this `setup()` method.

spflueger:
> > @redeboer Ok maybe we can do a quick call to discuss this.
> 
> @spflueger Done heavy_check_mark Usage can be previewed [here](https://tensorwaves--399.org.readthedocs.build/en/399/usage/basics.html#minuit2). Not the most pretty, but less error prone than this `setup()` method.

LGTM

PR #459, opened 2022-09-01
FEAT: support Python 3.10 and TF >2.6

redeboer:
This PR will also address [this problem](https://github.com/ComPWA/tensorwaves/runs/7944398788).


jonas-eschle:
phasespace 1.7 is out, Python 3.10 and TF 2.7+ is now supported

PR #476, opened 2023-01-23
FEAT: add `migrad_args` option to `Minuit2` optimizer

### LINKED ISSUE #475: Setting number of iterations in migrad
I noticed that is not possible to set the number of function calls currently.
This has to be set when calling `migrad`, currently `migrad` does not get any parameters which results in `Minuit` setting an max number of function calls:
https://iminuit.readthedocs.io/en/stable/reference.html?highlight=_migrad_maxcall#iminuit.Minuit.migrad

redeboer:
Closes #475

### DIFF
--- a/src/tensorwaves/optimizer/minuit.py
+++ b/src/tensorwaves/optimizer/minuit.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import logging
 import time
-from typing import Callable, Iterable, Mapping
+from typing import Any, Callable, Iterable, Mapping
 
 import iminuit
 from tqdm.auto import tqdm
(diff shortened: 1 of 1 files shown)


Leongrim:
Works really well.
Only one thing. If the input is not an `int` the error `Minuit` throws can be a little confusing. 

```
TypeError: __call__(): incompatible function arguments. The following argument types are supported:
    1. (self: iminuit._core.MnApplication, arg0: int, arg1: float) -> ROOT::Minuit2::FunctionMinimum

Invoked with: <iminuit._core.MnMigrad object at 0x7f62bc94a830>, 1000000.0, 0.1```

I`m not sure if this is something we should take into account or not. 
But since it already changing `ncall` is something that should be rarely necessary to change.

Leongrim:
I used a flaot (`1e6`).

PR #521, opened 2024-03-06
WIP: functional API

jonas-eschle:
If we swap the order of data and params, the params associated with the model are treated as defaults and can be overriden by giving them explicitly.

This (to my understanding) makes the function callable with parameters as args instead of updating the parameters in the model.

There should not be any legacy issue as I hardly doubt anyone gives _wrong_ parameter values and relies on the functionto overwrite it with the defaults?

If that's fine, I'll gladly add a quick test, changelog etc.

What do you thing @redeboer ?

### DIFF
--- a/src/tensorwaves/function/__init__.py
+++ b/src/tensorwaves/function/__init__.py
@@ -106,7 +106,7 @@ class ParametrizedBackendFunction(ParametrizedFunction):
         self.__parameters = dict(parameters)
 
     def __call__(self, data: DataSample) -> np.ndarray:
-        extended_data = {**data, **self.__parameters}  # type: ignore[arg-type]
+        extended_data = {**self.__parameters, **data}  # type: ignore[arg-type]
         return self.__function(extended_data)  # type: ignore[arg-type]
 
     @property


jonas-eschle:
Thanks for the explanations @redeboer, that's useful insights!

But I think my point is different: the _order_ of the arguments doesn't matter (as it's a mapping), but the code has a different meaning
```
params = {**p1, **p2}
```
gives precedence to p2

In the code, this means that it gives precedence to the default args. So think of the following

assuming it takes x (data) and mu, sigma (parameters). The model has a default for mu and sigma.

We can call it with just `data = {'x': array}`. That's fine, it uses mu & sigma as defaults.

But we cannot explicitly say the parameters here, i.e. `data = {'x': array, 'mu': float}` will _not_ use `mu`, because it will be override by the default.

Now it says basically "if the users gives a parameter value, override it and ignore it, use the default". On the other hand, if we swap as suggested, it's "use the default values if not given, but if the user provides a parameter _explicitly_, use that instead"

So nothing changes anywhere basically _except_ that a parameter can be given explicitly. Actually, the current order of _deliberatly_ (and silently!) overwriting parameters that are given explicitly to the function is probably not intentional? It looks a bit to me as if this always should have been the other way around?
Or is there maybe any disadvantage that I don't see?

jonas-eschle:
Yes, exactly, that's the idea, and I think, except of adding in the docs that you _can_ also use it directly, there is nothing to change, should be 100% backwards compatible but rather forward compatible (at least if I think about the best API idea).


> Btw, is there a speficic use case you have in mind?

The main consequence: it becomes jittable (right? Or am I missing something here?)! Also, the parametrized function becomes, automatically, from a very specific object that has a very specific interface (such as the parameter update) to a general function that takes the variables as inputs.  But even without the jitting, for example for minimization, it's now a completely functional API _aswell_, basically for free. So JIT and minimization is the main motivation. (I thought grads also work, that would have been better, but yep). 

So there is a difference between _calling the function with specific parameters_ and _changing the default values of the function_. While autodiff doesn't quite work with it (not sure why?), it's also the way to go for that.

> it's a bit hacky and going against the way the interfaces are designed now (the ParametrizedFunction call expects a DataSample that has arrays as values).

Is it? To me, it looks perfectly elegant. Especially as internally, the ParametrizedFunction does the exact same thing _but overrides the parameters if given_: It currently _doesn't allow_ but _forces it's state_. I think it would just solve, in a very elegant way, one of the largest problems with the API, namely that it isn't functional (but here I may miss details).

Since you're also working on jitting in other PRs I've seen, maybe I am really just missing details. But then, the functional API that it offers for free seems, without a downside I think, seems easy to enable?

jonas-eschle:
We can also discuss this tomorrow, it's an interesting issue ;)

>  If not, why would it become jittable if parameters can be passed to the call method?

It passes through a "symbolic parameter", much like sympy, but care must be taken that it's not converted somewhere. Also, it would not reflect any updated state, leading to a different behavior if run with JIT vs without (i.e. the Python object won't be updated if jitted, but it will if not jitted if the parameter is set).

I think it's jittable because it's simply a function now? Hm, `ParametrizedBackendFunction` okay! So that's the "executable", I've been a bit confused about what to call then and what not to, I'll try that! The `ParametrizedBackendFunction.function` is maybe what I was "looking for"

jonas-eschle:
> what the effect of those internal parameters is, even if the unpacking order is swapped

actually, here, only Python is happening: yes, these are "symbolic parameters", but only the ones that survive the dict merge survive the jit, so to say

> Hah, actually this touches on a larger issue :D

Sounds good, that is well aligned and I am really happy to see the extensive work on the conversion and careful backend handling, it's quite a tricky issue (so happy to make use of it!)

jonas-eschle:
@redeboer feel free to close or merge, whatever you think fits best, as discussed, it's not crucial and I'll use other ways to access the function.

PR #534, opened 2024-10-11
MAINT: drop Python 3.8 support due to EOL

grayson-helmholz:
Similarly to ComPWA/qrules#291 this PR will drop the
support for python3.8 as it has reached its EOL.
It removes then unneccessary constraints on the environment and older
syntax neccessary to be backward compatible.
Closes ComPWA/policy#395


redeboer (approved):
Perfect, thanks 🦖

